PR #446, opened 2019-11-01
【setHead mu】

516108736:


core/minorblockchain.go
```diff
@@ -311,16 +311,19 @@ func (m *MinorBlockChain) loadLastState() error {
 
 	return nil
 }
+func (m *MinorBlockChain) SetHead(head uint64) error {
```
pingke: empty line

core/minorblockchain.go
```diff
@@ -311,16 +311,19 @@ func (m *MinorBlockChain) loadLastState() error {
...
+	m.chainmu.Lock()
+	defer m.chainmu.Unlock()
+	return m.setHead(head)
+}
 
 // SetHead rewinds the local chain to a new head. In the case of Headers, everything
```
pingke: move those comment to public func

PR #449, opened 2019-11-02
  remove TD from root chain

pingke:


core/rootblockchain.go
```diff
@@ -522,12 +499,6 @@ func (bc *RootBlockChain) WriteBlockWithState(block *types.RootBlock) (status Wr
 		// the block generated by the local miner as the canonical block.
 		if block.NumberU64() < currentBlock.NumberU64() {
```
516108736: not need to check localTd==externTd

https://github.com/QuarkChain/pyquarkchain/blob/4014a79bc852e6e92d9939f235c66edbfc3a9a1a/quarkchain/cluster/root_state.py#L619

PR #454, opened 2019-11-05
Support xshard gas estimate

mask-pp:


core/minorblockchain_addon.go
```diff
@@ -1097,10 +1095,22 @@ func (m *MinorBlockChain) EstimateGas(tx *types.Transaction, fromAddress account
 	cap := hi
 
 	runTx := func(gas uint32) error {
+		evmTx, err := tx.CopyEvmTx()
```
pingke: why we need to copy?

mask-pp: don't need,there is a mistake.

core/types/transaction.go
```diff
@@ -198,10 +198,11 @@ func (tx *EvmTransaction) GasTokenID() uint64 {
 func (tx *EvmTransaction) TransferTokenID() uint64 {
 	return tx.data.TransferTokenID
 }
-func (tx *EvmTransaction) FromFullShardKey() uint32 { return tx.data.FromFullShardKey.GetValue() }
-func (tx *EvmTransaction) ToFullShardKey() uint32   { return tx.data.ToFullShardKey.GetValue() }
-func (tx *EvmTransaction) FromChainID() uint32      { return tx.data.FromFullShardKey.GetValue() >> 16 }
-func (tx *EvmTransaction) ToChainID() uint32        { return tx.data.ToFullShardKey.GetValue() >> 16 }
+func (tx *EvmTransaction) SetFromFullShardKey(key uint32) { k := Uint32(key); tx.data.FromFullShardKey = &k }
```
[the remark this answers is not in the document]
mask-pp: don't need.

PR #461, opened 2019-11-06
fix issue 430 400

tomdong2010:


internal/qkcapi/api.go
```diff
@@ -784,9 +784,14 @@ func (e *EthBlockChainAPI) GetBalance(address common.Address, fullShardKey *hexu
...
 }
+func (e *EthBlockChainAPI) GetTransactionCount(address common.Address, fullShardKey string) (hexutil.Uint64, error) {
 
-func (e *EthBlockChainAPI) GetTransactionCount(address common.Address, fullShardKey *hexutil.Uint) (hexutil.Uint64, error) {
-	fullShardId, err := getFullShardId(fullShardKey)
+	fullShard, err := strconv.ParseUint(fullShardKey, 0, 16)
+	if err != nil {
+		return 0, err
+	}
+	fullShardUnit := uint(fullShard)
```
pingke: clusterCfg.Quarkchain.GetFullShardIdByFullShardKey(uint32(fullShard))

internal/qkcapi/api.go
```diff
@@ -785,8 +785,13 @@ func (e *EthBlockChainAPI) GetBalance(address common.Address, fullShardKey *hexu
 	return (*hexutil.Big)(balance), nil
 }
 
-func (e *EthBlockChainAPI) GetTransactionCount(address common.Address, fullShardKey *hexutil.Uint) (hexutil.Uint64, error) {
-	fullShardId, err := getFullShardId(fullShardKey)
+func (e *EthBlockChainAPI) GetTransactionCount(address common.Address, fullShardKey string) (hexutil.Uint64, error) {
```
mask-pp: fullShardKey must be is optional, and the version is that hexutil.Uint64 don't support 0x00XX type.

tomdong2010: @mask-pp if change hexutil  ,i  agree

tomdong2010: ok

PR #466, opened 2019-11-07
fix sync bug

mask-pp:


frode88:
Hi.
didnt fix the issue.

Skjermbilde

PR #481, opened 2019-11-14
fix stats status & etc.

syntrust:


cluster/master/api_backend.go
```diff
@@ -390,10 +390,6 @@ func (s *QKCMasterBackend) GetTip() uint64 {
 	return s.rootBlockChain.CurrentBlock().NumberU64()
 }
 
-func (s *QKCMasterBackend) IsSyncIng() bool {
```
mask-pp: move IsSyncing(), IsMining(), CurrentBlock(), GetSlavePoolLen() funcs to api_backend.go in master forder

syntrust: done

PR #503, opened 2019-11-22
【log-display】modify log

516108736:


516108736:
@tomdong2010 please fix test err

PR #525, opened 2019-12-04
test qkcpow case

tomdong2010:


consensus/qkchash/qkchash_test.go
```diff
@@ -126,3 +126,29 @@ func sealBlock(t *testing.T, q *QKCHash, h *types.RootBlockHeader) {
...
+	err = q.VerifySeal(nil, header, big.NewInt(10))
+	assert.NoError(err, "should have correct nonce / mix digest")
+
+	// Wrong
+	header.Nonce = block.IHeader().GetNonce() - 1
+	err = q.VerifySeal(nil, header, big.NewInt(10))
+	assert.Error(err, "should have error because of the wrong nonce")
+	//wrong epoch
+	header.Nonce = block.IHeader().GetNonce()
+	err = q.VerifySeal(nil, header, big.NewInt(1000))
```
syntrust: why change diff?

tomdong2010: test VerifySeal  to show  err        maybe add instroduce about it

syntrust: it's about  epoch, so change height as python does

syntrust: impossible and meaningless to change height without change header. remove the epoch case.

PR #531, opened 2019-12-12
【gc】support archive and full mode

516108736:


core/rootblockchain.go
```diff
@@ -42,7 +42,7 @@ const (
 	receiptsCacheLimit        = 32
 	maxFutureBlocks           = 32
 	maxTimeFutureBlocks       = 30
-	triesInMemory             = 256
+	triesInMemory             = 128
```
pingke: still use 256?

core/minorblockchain_addon.go
```diff
@@ -406,11 +407,54 @@ func (m *MinorBlockChain) InitFromRootBlock(rBlock *types.RootBlock) error {
...
+func (m *MinorBlockChain) reRunBlockWithState(block *types.MinorBlock) error {
+	blockWithoutState := make([]types.IBlock, 0)
+	for {
+		if _, err := m.StateAt(block.Meta().Root); err == nil {
+			log.Info("reRunBlockWithState blockchain to past state", "number", block.Number(), "hash", block.Hash().String())
+			break
+		}
+		blockWithoutState = append(blockWithoutState, block)
+		block = m.GetMinorBlock(block.ParentHash())
+		if qkcCommon.IsNil(block) {
```
pingke: check block high == 0 ?

core/minorblockchain.go
```diff
@@ -282,9 +280,9 @@ func (m *MinorBlockChain) loadLastState() error {
 	if _, err := m.StateAt(currentBlock.GetMetaData().Root); err != nil {
 		// Dangling block without a state associated, init from scratch
 		log.Warn("Head state missing, repairing chain", "number", currentBlock.NumberU64(), "hash", currentBlock.Hash())
-		if err := m.repair(&currentBlock); err != nil {
-			return err
-		}
+		//if err := m.repair(&currentBlock); err != nil {
```
pingke: why remove?

516108736: will repair in InitFromRootBlock
will remove repair func

core/minorblockchain_addon.go
516108736: always have trie while block height=0

PR #550, opened 2020-02-10
【fix-bug】

516108736:
1.rootBlock not contain minor block
2.miner clinet timeout


516108736:
https://github.com/QuarkChain/goquarkchain/issues/560

PR #555, opened 2020-02-21
fix websocket logs

syntrust:


cluster/slave/api.go
```diff
@@ -132,7 +132,7 @@ func (api *PublicFilterAPI) NewHeads(ctx context.Context, fullShardId hexutil.Ui
 }
 
 // Logs creates a subscription that fires for all new log that match the given filter criteria.
-func (api *PublicFilterAPI) Logs(ctx context.Context, crit rpc.FilterQuery, fullShardId hexutil.Uint) (*rpc.Subscription, error) {
```
pingke: is this api same as py after change ?

cluster/slave/filters/filter_system.go
```diff
@@ -179,20 +179,11 @@ func (es *EventSystem) subscribe(sub *subscription) *Subscription {
...
-		return nil, err
-	}
-
-	if crit.FromBlock == nil || crit.FromBlock.Int64() == qrpc.LatestBlockNumber.Int64() {
-		crit.FromBlock = big.NewInt(int64(header.Number))
-	}
-	if crit.ToBlock == nil || crit.ToBlock.Int64() == qrpc.LatestBlockNumber.Int64() {
-		crit.FromBlock = big.NewInt(int64(header.Number))
+	if crit.ToBlock == nil {
+		crit.ToBlock = new(big.Int).SetInt64(qrpc.LatestBlockNumber.Int64())
```
pingke: please check this value -- qrpc.LatestBlockNumber.Int64(), is it -1?

cluster/slave/api.go
syntrust: this fix aim to be same with py

cluster/slave/filters/filter_system.go
syntrust: yes. it dynamicly points to the current block

PR #565, opened 2020-03-15
add checkdb

516108736:


qkcdb/database.go
```diff
@@ -31,25 +31,24 @@ func NewRDBDatabase(file string, clean, isReadOnly bool) (*RDBDatabase, error) {
...
-	opts.SetMaxTotalWalSize(uint64(cache * 1024 * 1024))
 	// sets the maximum number of write buffers that are built up in memory.
 	opts.SetMaxWriteBufferNumber(3)
 	// sets the target file size for compaction.
-	opts.SetTargetFileSizeBase(6710886)
+	opts.SetTargetFileSizeBase(67108864)
 	opts.SetCreateIfMissing(true)
-	opts.IncreaseParallelism(3)
-	opts.SetMaxBackgroundFlushes(1)
+	opts.SetMaxOpenFiles(100000)
```
pingke: why set it to 100000?

cmd/cluster/main.go
```diff
@@ -100,7 +101,7 @@ func init() {
 		//cache := ctx.GlobalInt(utils.CacheFlag.Name)
 		//gogc := math.Max(10, math.Min(90, 100*(float64(cache)/float64(allowance))))
 		//log.Debug("Sanitizing Go's GC trigger", "percent", int(gogc))
-		//godebug.SetGCPercent(int(gogc))
+		godebug.SetGCPercent(int(10))
```
pingke: make it as a param?

qkcdb/database.go
516108736: https://github.com/QuarkChain/pyquarkchain/blob/947fc4c10d0c60982b724e3e974738e2ad2aab05/quarkchain/db.py#L68   same as pyquarkchain

cmd/cluster/main.go
516108736: will modify : same as go-ethereum

.github/workflows/checkdb.yml
```diff
@@ -1,11 +1,12 @@
 name: checkdb
 on:
   schedule:
-    - cron: '0 13 * * *'
+    - cron: '1 8 * * *'
```
ninjaahhh: since pyquarkchain is already using 8:01 UTC, I suggest delaying ~5 hours so `0 13 * * *` is actually a good choice

.github/workflows/checkdb.yml
```diff
@@ -34,4 +35,13 @@ jobs:
           ./cluster --cluster_config ../../mainnet/singularity/cluster_config_template.json --service S3>> S3.log 2>&1 &
           sleep 3
           cd - && cd cmd/cluster
-          ./cluster --cluster_config ../../mainnet/singularity/cluster_config_template.json --check_db --check_db_rblock_to=400000
+
+          WEEK_DAY=$(date +%w)
+          date
+          if [ $WEEK_DAY -eq 5 ];then
```
ninjaahhh: as discussed offline, should be one day earlier, thus should equal 6?

.github/workflows/checkdb.yml
516108736: ok

.github/workflows/checkdb.yml
516108736: ok

cmd/cluster/main.go
```diff
@@ -97,10 +99,10 @@ func init() {
...
 		}
 		// Ensure Go's GC ignores the database cache for trigger percentage
-		//cache := ctx.GlobalInt(utils.CacheFlag.Name)
-		//gogc := math.Max(10, math.Min(90, 100*(float64(cache)/float64(allowance))))
-		//log.Debug("Sanitizing Go's GC trigger", "percent", int(gogc))
-		//godebug.SetGCPercent(int(gogc))
+		cache := ctx.GlobalInt(utils.CacheFlag.Name)
+		gogc := math.Max(10, math.Min(90, 100*(float64(cache)/float64(allowance))))
+		log.Info("Sanitizing Go's GC trigger", "cache", cache, "allowance", allowance, "percent", int(gogc))
+		godebug.SetGCPercent(int(gogc))
```
pingke: need to check how many slave and master run on that machine

PR #594, opened 2020-04-27
modify cluster sh

516108736:
1.add log info
2.chmod +x *.sh


cmd/cluster/run_cluster.sh
```diff
@@ -1,20 +1,28 @@
...
 done
 
 sleep 5s
 
 # start master
 cmd="./cluster --cluster_config "${configPath}" >>master.log 2>&1 &"
-echo $cmd
 eval $cmd
+pid=$!
+echo "Start master successfull pid="$pid" logFile=master.log"
```
pingke, line 28: master.log  path?

PR #608, opened 2020-06-19
Users/pingke/goqkchashx

pingke:
1. implement llrb for uint64
2. implement qkchashx using go
3. refactor qkchash consensus

after this change:
 building quarkchain without tags will use go qkchashx
 building quarkchain using tag nt will use native qkchash


consensus/qkchash/algorithm_native.go
```diff
@@ -0,0 +1,111 @@
...
+	if epoch < lenCaches {
+		return c.caches[epoch]
+	}
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	needAddCnt := epoch - lenCaches + cacheAheadRound
+	seed := c.caches[len(c.caches)-1].seed
+	for i := 0; i < needAddCnt; i++ {
+		seed = crypto.Keccak256(seed)
+		c.caches = append(c.caches, generateCache(cacheEntryCnt, seed))
```
ninjaahhh, line 50: so we never remove cache? should we trim the old entries (like LRU-style) avoid consuming too much memory?

consensus/qkchash/llrb.go
```diff
@@ -0,0 +1,457 @@
+package qkchash
+
+const (
+	Null_Value = uint64(0xFFFFFFFFFFFFFFFF)
+)
+
+// Tree is a Left-Leaning Red-Black (LLRB) implementation of 2-3 trees
+type LLRB struct {
```
ninjaahhh, line 8: nit: seems those data structures don't need be exported

consensus/qkchash/llrb.go
```diff
@@ -0,0 +1,457 @@
...
+
+func (t *LLRB) replaceOrInsert(h *Node, item uint64) (*Node, uint64) {
+	if h == nil {
+		return newNode(item), Null_Value
+	}
+
+	h = t.walkDownRot23(h)
+
+	var replaced uint64
+	if item < h.Value { // BUG
```
ninjaahhh, line 133: what does `BUG` mean?

consensus/qkchash/llrb.go
ninjaahhh: also, seems you are using https://github.com/petar/GoLLRB/blob/master/llrb/llrb.go, does its license allow this usage without attribution?

ninjaahhh:
also, please have a better name for this PR

consensus/qkchash/llrb.go
pingke: remove

consensus/qkchash/algorithm_native.go
pingke: One cache for every 30000 blocks, should not cache too much.

consensus/qkchash/llrb.go
pingke: updated, just keep functions we need, and those functions' implementation is similar to cpp implementation.

consensus/qkchash/algorithm_native.go
ninjaahhh: that's about 4~5 days. if we want to make the process long-lasting, should we at least trim the cache list a little bit?

consensus/qkchash/algorithm.go
```diff
@@ -45,18 +48,16 @@ func (c *cacheSeed) getCacheFromHeight(block uint64, useNative bool) qkcCache {
...
 	for i := 0; i < needAddCnt; i++ {
 		seed = crypto.Keccak256(seed)
-		c.caches = append(c.caches, generateCache(cacheEntryCnt, seed, useNative))
+		c.caches = append(c.caches, generateCache(cacheEntryCnt, seed))
 	}
 	return c.caches[epoch]
 }
 
 // qkcCache is the union type of cache for qkchash algo.
 // Note in Go impl, `nativeCache` will be empty.
```
qizhou, line 57: Comment is out-dated?

consensus/qkchash/algorithm_native.go
qizhou: Would this mean that we will cache for all epochs?  E.g., 3720294 miner blocks means 124 caches?  BTW: The code here seems to be duplicate of non-native version.

consensus/qkchash/algorithm.go
pingke: updated， remove those comments

consensus/qkchash/algorithm_native.go
pingke: 3720294 miner blocks mean 124 caches? yes. so 16k for one year.
The code here seems to be duplicate of non-native version. I have removed dup code.

pingke: after test, in one cache, ls use 1m, and nativeCache use 2m. add remove cache code.

ninjaahhh:
LGTM

consensus/qkchash/qkchash.go
```diff
@@ -2,21 +2,95 @@ package qkchash
...
+	for i := 0; i < needAddCnt; i++ {
+		seed = crypto.Keccak256(seed)
+		c.caches = append(c.caches, generateCache(cacheEntryCnt, seed))
+	}
+	if needAddCnt > 2*cacheAheadRound {
+		for i := 0; i < needAddCnt-2*cacheAheadRound; i++ {
+			shrinkCache(c.caches[i])
+		}
+	}
+	return c.caches[epoch]
```
qizhou, line 58: What happen if the epoch's cache is shrunk (i.e., ls == nil), i.e., a client wants to get a cache from an old epoch?

pingke: if the epoch is an old one, it will get from checkCache(c.caches[epoch]) which will fill in if the cache has been shrunk.
	if epoch < lenCaches {
		return checkCache(c.caches[epoch])
	}

qizhou: Then I believe there is a race condition between
1; after return c.caches[epoch] and before c.caches[epoch].ls is used; 2; the cache is evict, i.e., ls = nil.

consensus/qkchash/qkchash.go
```diff
@@ -2,21 +2,102 @@ package qkchash
...
+	cacheEntryCnt   = 1024 * 64
+	cacheAheadRound = 64 // 64*30000
 )
 
+var (
+	EpochLength = uint64(30000) //blocks pre epoch
+)
+
+type cacheSeed struct {
+	mu     sync.Mutex
```
qizhou, line 26: It will be great if a few comments about the protection scope of the mu is added.

PR #630, opened 2021-09-10
update EIP155 Contants

ping-ke:


qizhou (approved):
Should pyqkc needs to be updated accordingly?

PR #665, opened 2025-02-27
Fix Action failure

ping-ke:
The goquarkchain action [build_and_test](https://github.com/QuarkChain/goquarkchain/actions/runs/13541146085/job/37842295698#logs) which use quarkchaindocker/goquarkchain will fail with error 
`/__e/node20/bin/node: /lib/x86_64-linux-gnu/libc.so.6: version "GLIBC_2.28" not found (required by /__e/node20/bin/node)`. 
As the quarkchaindocker/goquarkchain is build on `ubuntu:18.04`, GLIBC_2.28 is not support. So update the docker using `ubuntu:20.04` and also update the rocksdb version from `v6.1.2` to `v6.20.3`.

**Changes**
1. Update the Dockfile using `ubuntu:20.04` and update rocksdb version from `v6.1.2` to `v6.20.3`.
2. Build docker and push as quarkchaindocker/goquarkchain:test1.6.1
3. Update actions/checkout using v4 and using docker quarkchaindocker/goquarkchain:test1.6.1 to fix action failure.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -6,7 +6,7 @@ jobs:
     container: quarkchaindocker/goquarkchain
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
       - name: daily-test
         run: |
           cd consensus/qkchash/native && make && cd -
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -5,14 +5,10 @@ on: [push]
 jobs:
   build_and_test:
     runs-on: ubuntu-latest
-    container: quarkchaindocker/goquarkchain:mainnet1.6.1
+    container: quarkchaindocker/goquarkchain:test1.6.1
 
     steps:
-      - name: upgrade
-        run: apt update && apt upgrade -y
-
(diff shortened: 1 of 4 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -1,14 +1,16 @@
 name: go-test
+
 on: [push]
+
 jobs:
   build_and_test:
     runs-on: ubuntu-latest
-    container: quarkchaindocker/goquarkchain
+    container: quarkchaindocker/goquarkchain:test1.6.1
```
syntrust, line 8: why labeld test1.6.1?

ping-ke: This is base on mainnet1.6.1 which is the latest one and upgrade the ubuntu to 20.04 for test, as we do not have an offical release so far.

docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:18.04
+FROM ubuntu:20.04
```
qzhodl, line 1: Do we want to update this according our meeting

qzhodl (approved):

syntrust (approved):

PR #668, opened 2025-03-26
Quarkchain: Root chain support POSW stake DecayByHeightAndTime

ping-ke:
related issue: https://github.com/QuarkChain/pm/issues/12

Added function and unit test for root chain to support PoSW stake decay by epoch. 
The enable time (ENABLE_ROOT_POSW_STAKING_DECAY_TIMESTAMP in config) currenlty set to 1746060000 which is 2025-05-01 08:40:00. And  this can be changed if we finalize.


core/rootblockchain.go
```diff
@@ -114,6 +120,29 @@ type RootBlockChain struct {
 	rootChainStakesFunc func(address account.Address, lastMinor common.Hash) (*big.Int, *account.Recipient, error)
 }
 
+// WriteStatus status of write
```
qizhou, line 123: What is the purpose of moving these functions in the same file?

ping-ke: It is a format change, just want to move all the RootBlockChain func together.

PR #670, opened 2025-04-07
security update to fix 'DoS via malicious p2p message'

ping-ke:
**Security issue**
[DoS via malicious p2p message](https://github.com/ethereum/go-ethereum/security/advisories/GHSA-ppjg-v974-84cm)

**Impact**
A vulnerable node, can be made to consume unbounded amounts of memory when handling specially crafted p2p messages sent from an attacker node.

**Details**
The p2p handler spawned a new goroutine to respond to ping requests. By flooding a node with ping requests, an unbounded number of goroutines can be created, leading to resource exhaustion and potentially crash due to OOM.

**Ethereum Patches** 
Fixed by https://github.com/ethereum/go-ethereum/pull/27887

### DIFF
--- a/p2p/peer.go
+++ b/p2p/peer.go
@@ -109,6 +109,7 @@ type Peer struct {
 	wg       sync.WaitGroup
 	protoErr chan error
 	closed   chan struct{}
+	pingRecv chan struct{}
 	disc     chan DiscReason
 
 	// events receives message send / receive events if set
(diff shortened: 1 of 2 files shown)


p2p/peer.go
```diff
@@ -201,8 +203,9 @@ func (p *Peer) run() (remoteRequested bool, err error) {
 		readErr    = make(chan error, 1)
 		reason     DiscReason // sent to the peer
 	)
-	p.wg.Add(1)
+	p.wg.Add(2)
 	go p.readLoop(readErr)
+	go p.pingLoop()
```
qzhodl, line 208: Which func call pingLoop() before this change?

ping-ke: no func all pingLoop before.

qzhodl (approved):
LGTM

PR #674, opened 2026-02-27
QuarkChain RPC health check workflow

syntrust:
Health check the RPC endpoints by querying `networkInfo` and verify the `networkId` of the result.

### DIFF
--- a/.github/workflows/rpc_healthcheck.yml
+++ b/.github/workflows/rpc_healthcheck.yml
@@ -51,8 +51,7 @@ jobs:
               -H 'Content-Type: application/json' \
               --data "$payload" \
               --output "$resp_file" --write-out '%{http_code}' \
-              "$RPC_URL" \
-              || echo '000'
+              "$RPC_URL" || true
           )"
 
           result="success"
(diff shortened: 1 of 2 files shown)


.github/workflows/rpc_healthcheck.yml
```diff
@@ -0,0 +1,251 @@
...
+  check-networkinfo:
+    name: Check ${{ matrix.network }} networkInfo
+    if: ${{ github.repository == 'QuarkChain/goquarkchain' }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - network: mainnet
```
qzhodl: There’s a list of RPCs that needs to be checked, and we don’t want to expose it to outsiders. Can we set the list through an environment variable?

syntrust: Added PRC URLs to the GitHub repo variable 'RPC_HEALTHCHECK_TARGETS_JSON', which can only be viewed by team members, and did some optimizations to the email report.

qzhodl (approved):

PR #678, opened 2026-05-19
fix: shard sync stuck after kill and restart — multi bugs in minor block sync pipeline

ping-ke:
## Summary

Fix a cascade of 5 independent bugs that together caused Shard 40001 to permanently stall on sync after an OOM/kill restart. The node's chain tip was stuck at 22774779 and could not insert blocks 22774780–22774782. Root cause analysis and full timeline are in `docs/incident/2026-05-18-shard40001-sync-unknown-ancestor.md`.

---

### Bug 1 — Typed-nil interface panic in `VerifyHeader` (`consensus/consensus.go`)

`GetBlock()` returns a `types.IBlock` interface. When the underlying `*MinorBlock` is nil, Go's interface value is **not** nil (it carries a non-nil type descriptor), so `parent == nil` evaluates `false` and `parent.NumberU64()` panics.

**Fix:** replace `parent == nil` with `qkccommon.IsNil(parent)`, which uses `reflect` to inspect the underlying pointer.

---

### Bug 2 — Sync batch loop exits early on known block (`cluster/shard/api_backend.go`)

When `InsertChainForDeposits` returns an empty xshard list (block already known), the loop executed `return nil` instead of `continue`, silently dropping all remaining blocks in the batch.

**Fix:** `return nil` → `continue`. Also distinguish the two sub-cases for `len(xshardLst) != 1`:
- Block committed via `insertSidechain` → `continue` (rest of batch unaffected)
- Chain shutting down (`procInterrupt`) → `return error` (force caller retry)

---

### Bug 3 — `isSameRootChain` violates precondition of `isSameChain` (`core/minorblockchain_addon.go`)

`isSameChain` requires `long.Number >= short.Number`. After a root-chain rewind, a minor block header's `prevRootBlockHash` can point to a root block taller than the current `rootTip`, causing `long < short` and a `log.Crit` crash.

**Fix:** add an early `return false` guard in `isSameRootChain` when `long.NumberU64() < short.NumberU64()`.

---

### Bug 4 — `HasBlock` checks commit marker only, not block body (`core/minorblockchain.go`)

The commit marker (`makeCommitMinorBlock` key) and block body (`blockKey`) are **independent** LevelDB keys. After an OOM/kill, the marker can persist while the body is lost. `HasBlock` returned `true`, so the block was filtered from the download list, but `GetBlock()` returned nil, producing `unknown ancestor`.

**Fix:** `HasBlock` now requires both `IsMinorBlockCommittedByHash(hash) && rawdb.HasBlock(m.db, hash)`.

---

#### `setHead` deletes block bodies, breaking crash recovery (`core/minorblockchain.go`)

The original `setHead` called `rawdb.DeleteMinorBlock` while rewinding, erasing block bodies. `reRunBlockWithState` replays transactions to rebuild the state trie and requires those bodies. Additionally, when the target block's state was missing, `setHead` reset `currentBlock`

The commit marker (`makeCommitMinorBlock` key) and block body (`blockKey`) are **independent** LevelDB keys. After an OOM/kill, the marker can persist while the body is
lost. `HasBlock` returned `true`, so the block was filtered from the download list, but `GetBlock()` returned nil, producing `unknown ancestor`.

**Fix:** `HasBlock` now requires both `IsMinorBlockCommittedByHash(hash) && rawdb.HasBlock(m.db, hash)`.

---

#### `setHead` deletes block bodies, breaking crash recovery (`core/minorblockchain.go`)

The original `setHead` called `rawdb.DeleteMinorBlock` while rewinding, erasing block bodies. `reRunBlockWithState` replays transactions to rebuild the state trie and
requires those bodies. Additionally, when the target block's state was missing, `setHead` reset `currentBlock` to genesis—undoing any state already rebuilt by `reRunBlockWithState`.

#### `setHead` deletes block bodies, breaking crash recovery (`core/minorblockchain.go`)

The original `setHead` called `rawdb.DeleteMinorBlock` while rewinding, erasing block bodies. `reRunBlockWithState` replays transactions to rebuild the state trie
and requires those bodies. Additionally, when the target block's state was missing, `setHead` reset `currentBlock` to genesis—undoing any state already rebuilt by
`reRunBlockWithState`.

**Fix:**
- Replace `rawdb.DeleteMinorBlock` with `rawdb.DeleteReceipts` + `rawdb.DeleteMinorBlockCommitStatus` (bodies preserved).
- Remove the state-missing → genesis reset; only reset to genesis when `currentBlock` is literally `nil`.

---

###  Bug 6 — Race condition: redundant CommitMinorBlockByHash outside all locks (cluster/shard/api_backend.go)

After InsertChainForDeposits completes (releasing m.chainmu), AddMinorBlock and AddBlockListForSync performed network calls and then called CommitMinorBlockByHash again — entirely outside any lock. This raced with AddRootBlock, which acquires m.chainmu to call setHead. The race window is the network-call latency (~10–100 ms).

When the race fires, setHead atomically deletes the block's body, commit marker, and canonical hash. The racy CommitMinorBlockByHash then re-writes the marker for a block whose body and canonical hash are now absent:

```
  Final state:  body[X]      = ABSENT
                marker[X]    = PRESENT  (re-written by racy call)
                canonHash[X] = ABSENT   (deleted by setHead)

  → HasBlock(X) = true  → sync skips X
  → GetBlock(X) = nil   → "unknown ancestor" for X+1

```
  The shard-layer calls were redundant: CommitMinorBlockByHash is already called inside insertChain (via WriteBlockWithState or insertSidechain) while m.chainmu is held.

**Fix:** remove both redundant CommitMinorBlockByHash calls from AddMinorBlock and AddBlockListForSync.

---

###  Bug 7 — Additional nil / typed-nil patterns analogous to Bug 1 (core/minorblockchain.go, core/minorblockchain_addon.go)

---

## Test Plan

- [x] Node survives OOM/kill restart and resumes sync without manual intervention
- [x] Shard with missing block body (marker present, body absent) re-downloads and inserts the block correctly
- [x] Unit test covered by `go test`
- [x] Node Sync to latest block and support mining request


docs/incident/2026-05-18-shard40001-sync-unknown-ancestor.md
```diff
@@ -0,0 +1,293 @@
+# Incident Report: Shard 40001 Sync Failure — "Unknown Ancestor" & Nil Pointer Panic
```
qzhodl, line 1: English should be used when opening a PR

cluster/shard/api_backend.go
```diff
@@ -430,6 +430,12 @@ func (s *ShardBackend) AddBlockListForSync(blockLst []*types.MinorBlock) error {
 		return nil
 	}
 
+	// Guard against being called before InitFromRootBlock completes.
```
qzhodl: This guard seems to deadlock the crash-recovery path.

In the incident scenario, `InitFromRootBlock()` can fail because `reRunBlockWithState()` cannot find a required minor block body. Since `ready` is only set to `true` after `InitFromRootBlock()` completes successfully, `IsInitialized()` remains false.

That means the subsequent root sync path cannot call `AddBlockListForSync()` to re-download and restore the missing block body, because this guard returns `shard not initialized yet` first. The node is then stuck in a loop:

`InitFromRootBlock` fails due to missing block body -> `IsInitialized() == false` -> `AddBlockListForSync` is rejected -> missing body is never restored -> next `InitFromRootBlock` fails again.

We probably need to distinguish “currently initializing” from “needs recovery / missing block body”, or allow the recovery sync path to download missing blocks before the shard is fully marked ready.

cmd/utils/flags.go
```diff
@@ -150,7 +150,6 @@ var (
 		Usage: "the batch size of root block check at the same time",
 		Value: 0,
 	}
-
```
qzhodl, line 153: Remove this change

core/minorblockchain.go
```diff
@@ -537,8 +538,12 @@ func (m *MinorBlockChain) Genesis() *types.MinorBlock {
 }
 
 // HasBlock checks if a block is fully present in the database or not.
+// A block is considered fully present only when both its commit marker and
+// its body exist in rawdb. A commit marker without a body can occur after a
+// crash (OOM/kill) where the marker persisted but the body was never flushed,
+// and would cause "unknown ancestor" errors in subsequent syncs.
 func (m *MinorBlockChain) HasBlock(hash common.Hash) bool {
-	return m.IsMinorBlockCommittedByHash(hash)
+	return m.IsMinorBlockCommittedByHash(hash) && rawdb.HasBlock(m.db, hash)
```
qzhodl, line 555: There is still a recovery gap for the “commit marker exists but block body is missing” case.

This PR updates `MinorBlockChain.HasBlock()` to require both the commit marker and the raw block body, so `SlaveBackend.AddBlockListForSync()` can correctly decide to re-download a missing block body.

However, after the block is downloaded, `ShardBackend.AddBlockListForSync()` still checks `getBlockCommitStatusByHash()` before inserting it. That function only checks `IsMinorBlockCommittedByHash()`, i.e. the commit marker. If the commit marker exists but the block body is missing, the shard treats the downloaded block as `BLOCK_COMMITTED` and skips it:

```go
if s.getBlockCommitStatusByHash(blockHash) == BLOCK_COMMITTED {
    continue
}

**Comment by syntrust** [comment]
Is there any indication that the node was stopped by an OOM or kill 9?  
I am curious why it took so long to discover the five bugs after they had been present for so long. Any idea?

File: core/minorblockchain_addon.go [hunk 5]
```diff
@@ -518,9 +522,16 @@ func (m *MinorBlockChain) InitFromRootBlock(rBlock *types.RootBlock) error {
 	err = m.reWriteBlockIndexTo(nil, block)
 	log.Info(m.logInfo, "init from root block end", m.CurrentBlock().NumberU64())
 	m.txPool = NewTxPool(DefaultTxPoolConfig, m)
+	if err == nil {
+		m.ready = true
```
qzhodl: `IsInitialized()` now returns the new `ready` flag, but `ready` is only set after `InitFromRootBlock()` completes successfully. Shard initialization can also complete through the genesis path (`InitGenesisState`) when `rBlock.Number() == genesisRootHeight`.

In that case the shard has a genesis block, `currentEvmState`, and tx pool initialized, but `ready` remains false. `AddBlockListForSync()` will then reject sync requests with `shard not initialized yet`.

We should either set `ready = true` after successful genesis initialization as well, or split the state into more precise phases instead of using `IsInitialized()` for only the `InitFromRootBlock` path.

core/minorblockchain.go
```diff
@@ -142,6 +142,7 @@ type MinorBlockChain struct {
 	lastDereferenceRoot      common.Hash
 	confirmedHeaderTip       *types.MinorBlock
 	initialized              bool
+	ready                    bool // set to true after InitFromRootBlock completes successfully
```
zhiqiangxu: `ready` is written in `InitFromRootBlock` (init goroutine) and read in `IsInitialized` from the gRPC `AddBlockListForSync` path on a different goroutine.

Use `atomic.Bool` or wrap with `m.mu`.

core/minorblockchain_addon.go
```diff
@@ -518,9 +522,16 @@ func (m *MinorBlockChain) InitFromRootBlock(rBlock *types.RootBlock) error {
...
+	}
 	return err
 }
 
+func (m *MinorBlockChain) IsInitialized() bool {
```
syntrust: Confusing name since we have an`initialized` field.

blockchaindevsh:
Worth mentioning that the latest go-ethereum fixes the partial write issue with two atomic batches:

```golang
  // Batch 1: block data (header + body + receipts + preimages atomic)                                                                                                                                                                              
  batch := bc.db.NewBatch()                                                                                                                                                                                                                       
  rawdb.WriteBlock(batch, block)         // body+header in one batch                                                                                                                                                                                
  rawdb.WriteReceipts(batch, block.Hash(), block.NumberU64(), receipts)                                                                                                                                                                             
  rawdb.WritePreimages(batch, statedb.Preimages())                                                                                                                                                                                                  
  batch.Write()    
```

```golang
// State trie commit happens after (separate persistence model with triedb) 
  batch := bc.db.NewBatch()                                                                                                                                                                                                                         
  rawdb.WriteHeadHeaderHash(batch, block.Hash())                                                                                                                                                                                                  
  rawdb.WriteHeadFastBlockHash(batch, block.Hash())                                                                                                                                                                                                 
  rawdb.WriteCanonicalHash(batch, block.Hash(), block.NumberU64())
  rawdb.WriteTxLookupEntriesByBlock(batch, block)                                                                                                                                                                                                   
  rawdb.WriteHeadBlockHash(batch, block.Hash())                                                                                                                                                                                                     
  batch.Write()

**Comment by ping-ke** [comment]
> Worth mentioning that the latest go-ethereum fixes the partial write issue with two atomic batches:
> 
> ```go
>   // Batch 1: block data (header + body + receipts + preimages atomic)                                                                                                                                                                              
>   batch := bc.db.NewBatch()                                                                                                                                                                                                                       
>   rawdb.WriteBlock(batch, block)         // body+header in one batch                                                                                                                                                                                
>   rawdb.WriteReceipts(batch, block.Hash(), block.NumberU64(), receipts)                                                                                                                                                                             
>   rawdb.WritePreimages(batch, statedb.Preimages())                                                                                                                                                                                                  
>   batch.Write()    
> ```
> 
> ```go
> // State trie commit happens after (separate persistence model with triedb) 
>   batch := bc.db.NewBatch()                                                                                                                                                                                                                         
>   rawdb.WriteHeadHeaderHash(batch, block.Hash())                                                                                                                                                                                                  
>   rawdb.WriteHeadFastBlockHash(batch, block.Hash())                                                                                                                                                                                                 
>   rawdb.WriteCanonicalHash(batch, block.Hash(), block.NumberU64())
>   rawdb.WriteTxLookupEntriesByBlock(batch, block)                                                                                                                                                                                                   
>   rawdb.WriteHeadBlockHash(batch, block.Hash())                                                                                                                                                                                                     
>   batch.Write()  
> ```

Yes, most database operations use `batch := m.db.NewBatch()`, and we can create a new PR to check if all places using `rawdb` need to be changed to batch.

File: core/minorblockchain.go [hunk 8]
```diff
@@ -525,8 +527,12 @@ func (m *MinorBlockChain) ExportN(w io.Writer, first uint64, last uint64) error
 // Note, this function assumes that the `mu` mutex is held!
 func (m *MinorBlockChain) insert(block *types.MinorBlock) {
 	// Add the block to the canonical chain number scheme and mark as the head
-	rawdb.WriteCanonicalHash(m.db, rawdb.ChainTypeMinor, block.Hash(), block.NumberU64())
-	rawdb.WriteHeadBlockHash(m.db, block.Hash())
+	batch := m.db.NewBatch()
+	rawdb.WriteCanonicalHash(batch, rawdb.ChainTypeMinor, block.Hash(), block.NumberU64())
+	rawdb.WriteHeadBlockHash(batch, block.Hash())
+	if err := batch.Write(); err != nil {
+		log.Error(m.logInfo, "Failed to write head block hash to database", "hash", block.Hash(), "number", block.NumberU64(), "err", err)
```
blockchaindevsh, line 538: change `log.Error` → `log.Crit`?

ping-ke: `insert()` is a normal forward block insertion operation—the block data has already been safely persisted before `insert()` is called. A write failure here simply means the head pointer still points to the previous block; it won't have a significant impact on the chain state.

blockchaindevsh: Worth mention that in the OLD code, the rawdb writers (`WriteCanonicalHash`, `WriteHeadBlockHash`, …) all do `log.Crit(...)` internally on `db.Put` failure.

The new change introduces a behavior change.

qzhodl:
Does PyQuarkChain have similar issues, or those are goquarkchain only bugs?

ping-ke:
## Comparison with PyQuarkChain

**TL;DR**: The bugs in this PR are **GoQuarkChain-only**, caused by the introduction of `setHead`. PyQuarkChain does not have an equivalent `setHead` mechanism and therefore cannot trigger these issues.

### 1. Why `setHead` exists in GoQuarkChain

GoQuarkChain (like Geth) introduced `setHead` as a maintenance operation for chain reorganization:

```go
// core/minorblockchain.go:317
func (m *MinorBlockChain) setHead(head uint64) error
```

When a new fork tip arrives at a shard, GoQuarkChain must:
1. **Rewind** the local chain to the common ancestor — `setHead` deletes block bodies, receipts, and commit markers for blocks above the new head
2. **Insert** the new fork — `InsertChain`

This is directly inspired by [Geth v1.8.20's `BlockChain.SetHead`](https://github.com/ethereum/go-ethereum/blob/v1.8.20/core/blockchain.go#L277-L281):
> *"Rewinds the local chain to a new head. In the case of headers, everything above the new head will be deleted and the new one set. In the case of blocks though, the head may be further rewound if block bodies are missing (non-archive nodes after a fast sync)."*

In production GoQuarkChain clusters, `setHead` is essential because:
- **Crash recovery**: After a crash, uncommitted blocks need to be replayed from the last confirmed state — `setHead` rewinds to that state first
- **Sync anomalies**: When a shard falls out of sync with its root chain, `setHead` rewinds it so the sync task can rebuild from a known good point
- **Peer reorg handling**: When a peer sends a higher-difficulty block, the shard must rewind before inserting the new chain

**The problem is not that `setHead` exists — it's needed.** The problem is the **race** between `setHead` and `reWriteBlockIndexTo` (which is called by `AddRootBlock` when processing a root block that contains a reorged minor block). 

### 2. How PyQuarkChain handles reorg differently

PyQuarkChain has **no `setHead`**. Its reorg mechanism is entirely different:

```python
# shard_state.py:829
def __rewrite_block_index_to(self, minor_block):
    """Find the common ancestor, rewrite index till minor_block"""

    # Only rewrites INDEXES — never deletes block bodies from DB
    for block in old_chain:
        self.db.remove_transaction_index_from_block(block)
        self.db.remove_minor_block_index(block)
        self.__add_transactions_from_block(block)  # return txs to txPool

    for block in new_chain:
        self.db.put_transaction_index_from_block(block)
        self.db.put_minor_block_index(block)
        self.__remove_transactions_from_block(block)  # remove from txPool

# shard_state.py:1399
def __update_tip(self, block, evm_state):
    self.__rewrite_block_index_to(block)
    self.header_tip = block.header   # just change the tip pointer
    self.meta_tip = block.meta
```

**Key differences:**

| Aspect | PyQuarkChain | GoQuarkChain (buggy) |
|--------|-------------|---------------------|
| Block body deletion on reorg | **Never** — blocks stay in DB forever | Yes — `DeleteMinorBlock` (fixed to only delete receipts) |
| Index operation | Rewrite tx→block and height→block indexes | Rewrite canonical hash + block index |
| Concurrency with root block processing | `add_root_block` checks block existence with `contain_minor_block_by_hash` — no lock-order issue | `reWriteBlockIndexTo` acquires `chainmu` while `setHead` also acquires it — lock-order race |
| State rollback | Re-executed block by block during `add_block` (EVM state is recomputed) | Rewound via `setHead`, then re-inserted via `InsertChain` |
| Crash recovery | Not needed — no committed-but-uninserted blocks | Needed — `commitBlock` markers trigger replay of uncommitted bodies after restart |

Since PyQuarkChain never deletes block bodies and never holds multiple locks in conflicting orders during reorg, the specific race conditions and "unknown ancestor" bugs described in this PR **cannot occur** in PyQuarkChain.

### Summary

PyQuarkChain avoids these by design — it does not delete block bodies and does not have the concurrent lock paths that make the race possible. The fixes in this PR should not be needed for PyQuarkChain.

core/minorblockchain.go
```diff
@@ -317,22 +317,24 @@ func (m *MinorBlockChain) SetHead(head uint64) error {
...
+	// HasBlock returns false and the blocks are re-downloaded if needed.
+	// Receipts are also dropped; they are regenerated on re-insertion.
 	batch := m.db.NewBatch()
 	for block := m.CurrentBlock(); block != nil && block.NumberU64() > head; block = m.CurrentBlock() {
-		rawdb.DeleteMinorBlock(batch, block.Hash())
+		rawdb.DeleteReceipts(batch, block.Hash())
+		rawdb.DeleteMinorBlockCommitStatus(batch, block.Hash())
 		rawdb.DeleteCanonicalHash(batch, rawdb.ChainTypeMinor, block.NumberU64())
-		m.currentBlock.Store(m.GetBlock(block.ParentHash()))
+		m.currentBlock.Store(m.GetMinorBlock(block.ParentHash()))
```
syntrust: Would it be better to check for nil before store `m.GetMinorBlock(block.ParentHash())`?

ping-ke:
> Is there any indication that the node was stopped by an OOM or kill 9? I am curious why it took so long to discover the five bugs after they had been present for so long. Any idea?

On the crash cause: The node wasn't stopped by OOM or kill -9 — it crashed due to Bug 2 (typed-nil panic in VerifyHeader). GetBlock(22774780) returned a typed-nil types.IBlock (non-nil interface wrapping nil pointer), the parent == nil check passed, then parent.NumberU64() panicked. The crash repeated on every restart because Bug 1 created the inconsistent DB state.

On why it took so long: The trigger was Bug 1, a race condition that produces an inconsistent DB state (marker present, body absent). This race window is ~10–100ms and requires AddRootBlock to arrive precisely during the network call window (see [Race interleaving (for a single block X):](https://github.com/QuarkChain/goquarkchain/blob/fix/sync-unknown-ancestor/docs/incident/2026-05-18-shard40001-sync-unknown-ancestor.md#bug-1-race-condition--redundant-commitminorblockbyhash-outside-all-locks)) — extremely rare.

  All other bugs are harmless without Bug 1's inconsistent state. In normal operation (no race), each bug individually is either self-healing (sync retries recover) or never triggered. It took the rare race to create the preconditions for the others to surface, which is why the cascade took so long to manifest.

syntrust:
> > Is there any indication that the node was stopped by an OOM or kill 9? I am curious why it took so long to discover the five bugs after they had been present for so long. Any idea?
> 
> On the crash cause: The node wasn't stopped by OOM or kill -9 — it crashed due to Bug 2 (typed-nil panic in VerifyHeader). GetBlock(22774780) returned a typed-nil types.IBlock (non-nil interface wrapping nil pointer), the parent == nil check passed, then parent.NumberU64() panicked. The crash repeated on every restart because Bug 1 created the inconsistent DB state.
> 
> On why it took so long: The trigger was Bug 1, a race condition that produces an inconsistent DB state (marker present, body absent). This race window is ~10–100ms and requires AddRootBlock to arrive precisely during the network call window (see [Race interleaving (for a single block X):](https://github.com/QuarkChain/goquarkchain/blob/fix/sync-unknown-ancestor/docs/incident/2026-05-18-shard40001-sync-unknown-ancestor.md#bug-1-race-condition--redundant-commitminorblockbyhash-outside-all-locks)) — extremely rare.
> 
> All other bugs are harmless without Bug 1's inconsistent state. In normal operation (no race), each bug individually is either self-healing (sync retries recover) or never triggered. It took the rare race to create the preconditions for the others to surface, which is why the cascade took so long to manifest.

If so, the 'OOM/kill' mentioned in the PR summary is confusing.

ping-ke:
# SetHead Usage Scenarios in go-ethereum v1.8.20, Master Branch, and goquarkchain

## What is SetHead?

`SetHead()` is the blockchain rewind mechanism used to roll the local chain back to a specified block height and remove all chain data above that point.

Conceptually, SetHead is not part of the normal synchronization path. Instead, it is used for:

- Database recovery
- Chain reorganization (reorg) handling
- Chain configuration upgrades
- Genesis reset
- Manual debugging and maintenance

---

# 1. go-ethereum Master Branch

The modern Geth architecture (PoS + Snap Sync + Ancient Store) significantly expands the responsibilities of `SetHead()` compared to earlier releases.

---

## 1. Ancient Store Consistency Recovery

### File

```go
core/blockchain.go:528
```

### Scenario

During blockchain initialization, Geth verifies consistency among:

- Current Header
- Current Snap Block
- Ancient Store (Freezer)

If the chain head exceeds the actual frozen data stored in the freezer, it usually indicates:

- Crash during freezer flush
- Marker written successfully but freezer data not fully persisted

### Invocation

```go
if needRewind {
    log.Error("Truncating ancient chain")
    bc.SetHead(low)
}
```

### Purpose

Rollback the chain to:

```text
frozen limit
```

and restore consistency between the hot database and ancient store.

---

## 2. Chain Configuration Compatibility Rewind

### File

```go
core/blockchain.go:561
```

### Scenario

At startup, Geth checks whether the stored chain configuration matches the current configuration:

```go
rawdb.CheckDatabaseCompatibility()
```

Typical incompatibilities include:

- Modified fork block numbers
- Changed Shanghai activation timestamp
- Changed Cancun activation timestamp

### Invocation

```go
bc.SetHead(compatErr.RewindToBlock)
```

or

```go
bc.SetHeadWithTimestamp(...)
```

### Purpose

Rollback to a block before the incompatible fork activation point.

This prevents consensus divergence caused by configuration changes.

---

## 3. ResetWithGenesisBlock

### File

```go
core/blockchain.go:1212
```

### Scenario

Complete blockchain reset.

### Invocation

```go
bc.SetHead(0)
```

### Purpose

Return the chain to:

```text
Genesis Block
```

and rewrite genesis state from scratch.

---

## 4. Downloader Reorg Recovery

### File

```go
eth/downloader/downloader.go:576
```

### Scenario

During Snap Sync or Beacon Sync, a peer may report a common ancestor (`origin`) that is older than the locally frozen chain segment.

Condition:

```go
origin + 1 < frozen
```

This indicates:

```text
The local node previously synchronized further ahead,
but the network later experienced a reorganization.
```

As a result, some blocks stored in the ancient store are no longer canonical.

### Invocation

```go
d.blockchain.SetHead(origin)
```

### Purpose

Delete all ancient blocks between:

```text
origin + 1 ~ frozen - 1
```

and realign the local chain with network consensus.

---

### Meaning of origin and frozen

| Variable | Meaning |
|-----------|----------|
| origin | Common ancestor returned by consensus |
| frozen | Number of blocks already stored in freezer |
| frozen - 1 | Highest block number currently in freezer |

Example:

```text
frozen = 1000
origin = 950
```

Blocks:

```text
951 ~ 999
```

are no longer canonical and must be removed.

Executing:

```go
SetHead(950)
```

truncates those obsolete blocks.

---

## 5. Debug RPC

### Call Path

```text
debug_setHead
  ↓
internal/ethapi/api.go
  ↓
eth/api_backend.go
  ↓
blockchain.SetHead
```

### Scenario

Manual rollback initiated by an operator.

### Invocation

```go
func (b *EthAPIBackend) SetHead(number uint64) error {
    b.eth.handler.downloader.Cancel()
    return b.eth.blockchain.SetHead(number)
}
```

### Purpose

Rollback to a user-specified block height.

---

## Summary: Geth Master

| Scenario | File | Trigger | Target |
|-----------|--------|--------|--------|
| Ancient Store Recovery | blockchain.go:528 | Startup consistency check | Frozen limit |
| Config Compatibility | blockchain.go:561 | Fork configuration change | Before fork |
| Genesis Reset | blockchain.go:1212 | Chain reset | Block 0 |
| Downloader Reorg | downloader.go:576 | Ancient-store reorg | Common ancestor |
| Debug RPC | api_backend.go | Manual invocation | User-specified block |

---

# 2. go-ethereum v1.8.20

Geth v1.8.20 belongs to the:

```text
PoW Era
Fast Sync
No Snap Sync
No Beacon Chain
```

Therefore, SetHead usage is significantly simpler.

---

## 1. Startup Bad-Hash Recovery

### File

```go
core/blockchain.go:199
```

### Scenario

During blockchain initialization:

```go
VerifyHeader()
```

fails because a block header hash is invalid.

If the header still exists on the canonical chain, it implies:

```text
Database corruption
```

### Invocation

```go
bc.SetHead(header.Number.Uint64() - 1)
```

### Purpose

Rollback one block and resume normal operation.

---

## 2. Chain Configuration Compatibility Rewind

### File

```go
eth/backend.go:166
```

### Scenario

Startup detects:

```go
ConfigCompatError
```

due to mismatched chain configuration.

### Invocation

```go
eth.blockchain.SetHead(compat.RewindTo)
```

### Purpose

Rollback before the conflicting fork activation point.

---

## 3. ResetWithGenesisBlock

### File

```go
core/blockchain.go:420
```

### Invocation

```go
SetHead(0)
```

### Purpose

Reset the entire blockchain to genesis.

---

## 4. Debug RPC

### Call Path

```text
debug_setHead
 ↓
internal/ethapi/api.go
 ↓
eth/api_backend.go
 ↓
blockchain.SetHead
```

### Invocation

```go
b.protocolManager.downloader.Cancel()
b.blockchain.SetHead(number)
```

### Purpose

Manual rollback for debugging and maintenance.

---

## Summary: Geth v1.8.20

| Scenario | File |
|-----------|--------|
| Bad Hash Recovery | core/blockchain.go:199 |
| Config Compatibility | eth/backend.go:166 |
| Genesis Reset | core/blockchain.go:420 |
| Debug RPC | api.go → api_backend.go |

---

# 3. goquarkchain (MinorBlockChain.SetHead)

In QuarkChain, `MinorBlockChain.SetHead()` is used for shard-chain rollback and state recovery.

Compared with Geth, it includes additional business-specific scenarios such as:

- Transaction index rebuilding
- Broadcast failure compensation

---

## 1. Tx Index Rebuild During Reorg

### File

```go
core/minorblockchain.go:1428
```

### Scenario

When:

```go
commonBlock == nil
```

the old chain and new chain are actually the same chain, and only transaction indices need rebuilding.

### Invocation

```go
m.setHead(newBlockNumber)
```

### Purpose

Rollback to the target block and rebuild transaction indexes.

---

## 2. Rollback Protection After Broadcast Failure

### File

```go
cluster/shard/api_backend.go
```

### Scenario

After a new minor block is added successfully, QuarkChain performs several follow-up broadcasts:

1. Broadcast XShard transactions
2. Update shard statistics
3. Send minor block header to master chain

If any step fails, the node must rollback.

### Invocation

```go
setHead(currHead.Number)
```

Internally:

```go
s.MinorBlockChain.SetHead(head)
```

### Purpose

Undo the newly inserted block and restore state consistency.

---

## 3. ResetWithGenesisBlock

### File

```go
core/minorblockchain.go:470
```

### Invocation

```go
SetHead(0)
```

### Purpose

Reset the shard chain back to genesis.

---

## Summary: goquarkchain

| Scenario | File | Purpose |
|-----------|--------|--------|
| Tx Index Rebuild | minorblockchain.go:1428 | Rebuild indexes |
| Broadcast Failure Recovery | shard/api_backend.go | Undo inserted block |
| Genesis Reset | minorblockchain.go:470 | Return to genesis |

PR #680, opened 2026-06-08
fix(Bug2): Fix typed-nil interface panics in block chain traversal

ping-ke:
###  Root Cause

  Go interfaces carry two fields: a type pointer and a value pointer. When GetBlock() returns a (*types.MinorBlock)(nil) wrapped in a types.IBlock interface, the interface itself is non-nil (type field is set), so parent == nil evaluates to false. Any subsequent method call (e.g. parent.NumberU64()) dereferences a nil pointer and panics.

  This pattern appears in several places where blocks are looked up by hash and the result is tested with == nil.

###  Changes

  consensus/consensus.go — VerifyHeader
  - Replace parent == nil with qkccommon.IsNil(parent) to correctly detect a typed-nil IBlock returned when the parent body is absent from rawdb.
  - Without this fix, inserting a block whose parent was evicted/crashed triggers a nil-pointer panic instead of returning ErrUnknownAncestor.

  core/minorblockchain.go — setHead / reorg
  - setHead: replace m.GetBlock(...) with m.GetMinorBlock(...) so the stored value is a concrete *types.MinorBlock, eliminating an implicit typed-nil through the atomic.Value.
  - reorg: replace oldBlock != nil / newBlock != nil loop guards with !qkcCommon.IsNil(...) to correctly terminate traversal when a block is missing.

  core/minorblockchain_addon.go — reRunBlockWithState / AddRootBlock
  - reRunBlockWithState: capture parentHash before the GetMinorBlock call so the error message can reference the hash even when the returned pointer is nil (the old code dereferenced block after confirming it was nil).
  - AddRootBlock: use GetMinorBlock instead of GetBlock(...).(*types.MinorBlock) and guard the Store call so a missing tip block does not store a nil value.

  core/minorblockchain_test.go
  - Add TestVerifyHeaderTypedNilParentReturnsErrUnknownAncestor: inserts a block body into rawdb, deletes only the body (leaving commit status intact via rawdb.DeleteBlock), then asserts that ValidateBlock on a child returns consensus.ErrUnknownAncestor instead of panicking.

###  Test
`
  go test ./...
`

### DIFF
--- a/consensus/consensus.go
+++ b/consensus/consensus.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/QuarkChain/goquarkchain/account"
 	"github.com/QuarkChain/goquarkchain/cluster/config"
+	qkccommon "github.com/QuarkChain/goquarkchain/common"
 	"github.com/QuarkChain/goquarkchain/consensus/posw"
 	"github.com/QuarkChain/goquarkchain/core/types"
 	"github.com/ethereum/go-ethereum/common"
@@ -147,7 +148,7 @@ func (c *CommonEngine) VerifyHeader(
 		}
 	*/
 	parent := chain.GetBlock(header.GetParentHash())
-	if parent == nil {
+	if qkccommon.IsNil(parent) {
 		return ErrUnknownAncestor
 	}
 	if parent.NumberU64() != number-1 {
--- a/core/minorblockchain.go
+++ b/core/minorblockchain.go
@@ -322,7 +322,7 @@ func (m *MinorBlockChain) setHead(head uint64) error {
 	for block := m.CurrentBlock(); block != nil && block.NumberU64() > head; block = m.CurrentBlock() {
 		rawdb.DeleteMinorBlock(batch, block.Hash())
 		rawdb.DeleteCanonicalHash(batch, rawdb.ChainTypeMinor, block.NumberU64())
-		m.currentBlock.Store(m.GetBlock(block.ParentHash()))
+		m.currentBlock.Store(m.GetMinorBlock(block.ParentHash()))
 	}
 	batch.Write()
 	if currentBlock := m.CurrentBlock(); currentBlock != nil {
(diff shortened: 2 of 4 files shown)


iteyelmp (dismissed):

blockchaindevsh:
Verdict: approve the production code, fix the test and one missed site before merge.                                                                                                                                                              
  
  Holds up                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                  
  - consensus/consensus.go:151 (VerifyHeader) — typed-nil panic on parent.NumberU64(); qkccommon.IsNil is correct.                                                                                                                                  
  - core/minorblockchain.go:325 (setHead) — GetBlock → GetMinorBlock swap avoids storing typed-nil into atomic.Value; dynamic type matches other writers.
  - core/minorblockchain.go:1361,1367 (reorg loops) — brings these in line with the rest of the function which already uses IsNil.                                                                                                                  
  - core/minorblockchain_addon.go:542-545 (reRunBlockWithState) — the error string itself was dereferencing nil; captures parentHash before the lookup.                                                                                             
  - core/minorblockchain_addon.go:1090-1097 (AddRootBlock tip update) — single GetMinorBlock + nil-guard; removes typed-nil store into currentBlock.                                                                                                
                                                                                                                                                                                                                                                    
  Pre-merge asks                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                  
  1. The new test doesn't exercise the fix. core/minorblockchain_test.go:1366-1402 uses consensus.FakeEngine, whose VerifyHeader is a no-op (consensus/fakeengine.go:35-41). The test then routes through MinorBlockValidator.ValidateBlock and     
  short-circuits at lines 93-96 (!HasBlockAndState && !HasBlock → ErrUnknownAncestor). Passes with or without the line-151 fix. Use a real CommonEngine-derived engine, or call blockchain.engine.VerifyHeader(...) directly; rename the test
  accordingly.                                                                                                                                                                                                                                      
  2. Same pattern unfixed in Rollback at core/minorblockchain.go:737-741: m.GetBlock(currentBlock.ParentHash()) stored into atomic.Value then newBlock.Hash() called on the next line. Identical risk as the setHead fix.                         
                                                                                                                                                                                                                                                    
  Nits
                                                                                                                                                                                                                                                    
  - Squash second commit 8007a02 ("fix") into the main one.                                                                                                                                                                                         
  - Error msg at _addon.go:544 "missing block with parent hash %x" drops the child context (number + hash) available before the reassignment.
  - .github/workflows/rpc_healthcheck.yml in the local diff is PR #679 rebase noise; folds away on rebase.

qzhodl (dismissed):
LGTM once Alan's comment is resolved

qzhodl (dismissed):

qzhodl (dismissed):

iteyelmp (dismissed):

iteyelmp (approved):

qzhodl (approved):

PR #681, opened 2026-06-08
fix(Bug4): guard `isSameRootChain` against `log.Crit` crash when long < short

ping-ke:
### Problem

  `isSameChain` (the internal helper called by `isSameRootChain`) assumes its first argument (`long`) always has a block number ≥ the second argument (`short`). When that invariant is violated it calls `log.Crit`, which terminates the process.

  `isSameRootChain` is called as:
 ` isSameRootChain(m.rootTip, ancestorRootHeader)`

  During startup or after a root chain rewind, a minor block's `PrevRootBlockHash` can point to a root block taller than the current `m.rootTip`. This makes `long.Number < short.Number`, violating the invariant and causing a crash instead of a graceful false return.

 ### Fix

  Add a guard at the top of `isSameRootChain` (`core/minorblockchain_addon.go`):
```
  if long.NumberU64() < short.NumberU64() {
      return false
  }
```
  This short-circuits before reaching `isSameChain` and returns false cleanly, matching the semantically correct answer (a shorter chain cannot contain the longer one).

###  Test

  TestIsSameRootChainInvertedArgs (core/shardstate_test.go) covers two cases:
 | Case           | long           | short          | Expected             |
  |----------------|----------------|----------------|----------------------|
  | Normal order   | rb2 (height 2) | rb1 (height 1) | true (direct parent) |
  | Inverted order | rb1 (height 1) | rb2 (height 2) | false (no crash)     |

  Without the guard the inverted-order call hits `log.Crit` and exits the process. With the guard it returns false.

### DIFF
--- a/core/minorblockchain_addon.go
+++ b/core/minorblockchain_addon.go
@@ -377,6 +377,9 @@ func (m *MinorBlockChain) GetTransactionCount(recipient account.Recipient, hash 
 }
 
 func (m *MinorBlockChain) isSameRootChain(long types.IBlock, short types.IBlock) bool {
+	if long.NumberU64() < short.NumberU64() {
+		return false
+	}
 	f := func(hash common.Hash) common.Hash {
 		if b := m.GetRootBlockByHash(hash); b == nil {
 			return common.Hash{}
--- a/core/minorblockchain_addon.go
+++ b/core/minorblockchain_addon.go
@@ -1079,7 +1079,14 @@ func (m *MinorBlockChain) AddRootBlock(rBlock *types.RootBlock) (bool, error) {
 		return false, nil
 	}
 
+	// Hold m.mu for the entire root/minor head transition to prevent CreateBlockToMine
+	// from reading an inconsistent snapshot (new rootTip with stale currentBlock).
+	// Previously, the lock was released after updating rootTip but before rewinding
+	// currentBlock, allowing mining to observe the intermediate state and call
+	// isSameRootChain with mismatched heights.
 	m.mu.Lock()
+	defer m.mu.Unlock()
+
 	m.rootTip = rBlock
 	if shardHeader == nil {
 		m.confirmedHeaderTip = nil
@@ -1087,7 +1094,6 @@ func (m *MinorBlockChain) AddRootBlock(rBlock *types.RootBlock) (bool, error) {
 		m.confirmedHeaderTip = m.GetMinorBlock(shardHeader.Hash())
 	}
 
-	m.mu.Unlock()
 	origHeaderTip := m.CurrentBlock()
 	if shardHeader != nil {
 		origBlock := m.GetBlockByNumber(shardHeader.Number)
--- a/core/minorblockchain_addon.go
+++ b/core/minorblockchain_addon.go
@@ -1079,14 +1079,7 @@ func (m *MinorBlockChain) AddRootBlock(rBlock *types.RootBlock) (bool, error) {
 		return false, nil
 	}
 
-	// Hold m.mu for the entire root/minor head transition to prevent CreateBlockToMine
-	// from reading an inconsistent snapshot (new rootTip with stale currentBlock).
-	// Previously, the lock was released after updating rootTip but before rewinding
-	// currentBlock, allowing mining to observe the intermediate state and call
-	// isSameRootChain with mismatched heights.
 	m.mu.Lock()
-	defer m.mu.Unlock()
-
 	m.rootTip = rBlock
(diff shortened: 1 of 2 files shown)


blockchaindevsh:
Verdict: approve.

  Holds up

  - core/common.go:18 confirms isSameChain does log.Crit (process exit) on long < short. Guard at _addon.go:379-382 short-circuits before that.
  - All six call sites of isSameRootChain traced; return false semantically correct at each. Critical case is AddRootBlock's rewind loop at _addon.go:1099 — the old log.Crit aborted the process, the fix lets the loop keep rewinding.
  - Equal-height case unchanged (falls through to in-memory hash compare at common.go:20-22).                                                                                                                                                       
  - Test TestIsSameRootChainInvertedArgs (core/shardstate_test.go:3640-3673) covers both directions.                                                                                                                                                
                                                                                                                                                                                                                                                    
  Follow-up (out of scope for this PR)                                                                                                                                                                                                              
                                                                                                                                                                                                                                                    
  - core/rootblock_validator.go:129 calls v.blockChain.isSameChain(rootBlock, v.blockChain.GetBlock(key)) with no ordering guard. Same crash class on the root chain.                                                                               
  - core/minorblock_validator.go:192 calls isSameChain(GetParentHashByHash, prevBlock, prevConfirmedMinorHeader) without explicit minor-block ordering pre-check. Probably guaranteed by protocol invariant, but worth defense-in-depth.          
                                                                                                                                                                                                                                                    
  Nits
                                                                                                                                                                                                                                                    
  - One-line comment above the guard explaining why this isn't an invariant violation (legitimate after root-chain rewind) would help next reader.                                                                                                  
  - Test uses raw if result != false; surrounding tests use assert.False. Stylistic.
  - Commit message imprecise: claims reRunBlockWithState calls isSameRootChain directly — actually downstream via InsertChain → updateTip.                                                                                                          
  - Rebase folds away the .github/workflows/rpc_healthcheck.yml noise.

core/minorblockchain_addon.go
```diff
@@ -377,6 +377,9 @@ func (m *MinorBlockChain) GetTransactionCount(recipient account.Recipient, hash
 }
 
 func (m *MinorBlockChain) isSameRootChain(long types.IBlock, short types.IBlock) bool {
+	if long.NumberU64() < short.NumberU64() {
```
qzhodl, line 380: This fix looks correct as a crash guard: `CreateBlockToMine`/recovery paths can observe `rootTip.Number < current minor tip's PrevRootBlock` during root-chain rewind/startup, and `isSameRootChain` should return `false` rather than letting `isSameChain` hit `log.Crit`.

One follow-up concern: this also suggests `AddRootBlock` exposes a transient inconsistent head state. It updates `m.rootTip`, releases `m.mu`, and only later rewinds/updates `currentBlock`, while `CreateBlockToMine` reads both without taking the same lock. The guard prevents bad blocks/crashes, but a cleaner fix would make the root/minor head transition atomic or have mining read a consistent snapshot.

syntrust:
I believe the same issue may affect other 'isSameChain' callers, but it is acceptable to have follow-ups to track in addition to the fix in this PR.

syntrust (dismissed):

ping-ke:
Add issue https://github.com/QuarkChain/goquarkchain/issues/688 to follow up.

> I believe the same issue may affect other 'isSameChain' callers, but it is acceptable to have follow-ups to track in addition to the fix in this PR.

core/minorblockchain_addon.go
```diff
@@ -1076,15 +1079,21 @@ func (m *MinorBlockChain) AddRootBlock(rBlock *types.RootBlock) (bool, error) {
 		return false, nil
 	}
 
+	// Hold m.mu for the entire root/minor head transition to prevent CreateBlockToMine
+	// from reading an inconsistent snapshot (new rootTip with stale currentBlock).
+	// Previously, the lock was released after updating rootTip but before rewinding
+	// currentBlock, allowing mining to observe the intermediate state and call
+	// isSameRootChain with mismatched heights.
 	m.mu.Lock()
+	defer m.mu.Unlock()
```
qzhodl: But CreateBlockToMine does not require this lock

ping-ke: The change move to PR https://github.com/QuarkChain/goquarkchain/pull/689.

core/minorblockchain_addon.go
ping-ke: Resulve it https://github.com/QuarkChain/goquarkchain/pull/689.

PR #682, opened 2026-06-09
fix(Bug5): HasBlock requires block body in addition to commit marker

ping-ke:
### Problem

  HasBlock() previously only checked IsMinorBlockCommittedByHash() — the commit marker written to RocksDB at the end of InsertChain. After an kill or forced termination, RocksDB could flush the commit marker to disk while the block body may be deleted.

  On the next startup:
  - HasBlock(hash) returned true (marker present)
  - Sync assumed the block was already stored and skipped re-downloading it
  - The next block then failed with "unknown ancestor" or crash permanently, stalling the node

###  Root Cause

```
  // Before: marker-only check — blind to missing body
  func (m *MinorBlockChain) HasBlock(hash common.Hash) bool {
      return m.IsMinorBlockCommittedByHash(hash)
  }
```

###  Fix

  Require both the commit marker and the block body to exist:

```
  func (m *MinorBlockChain) HasBlock(hash common.Hash) bool {
      return m.IsMinorBlockCommittedByHash(hash) && rawdb.HasBlock(m.db, hash)
  }
```

  This mirrors the contract expected by the sync layer: a block is only "fully present" when its body can actually be retrieved.

###  Test
```
  go test ./...
```

### DIFF
--- a/cluster/shard/shard.go
+++ b/cluster/shard/shard.go
@@ -207,18 +207,12 @@ func (s *ShardBackend) initGenesisState(rootBlock *types.RootBlock) error {
 }
 
 func (s *ShardBackend) getBlockCommitStatusByHash(blockHash common.Hash) BlockCommitCode {
-	// If the block is committed, it means
-	// - All neighbor shards/slaves receives x-shard tx list
-	// - The block header is sent to master
-	// then return immediately
-	if s.MinorBlockChain.IsMinorBlockCommittedByHash(blockHash) {
+	// Require both the commit marker and the block body. A crash can leave the
+	// marker present but the body absent; in that case the block must be
+	// re-inserted so the body is restored.
+	if s.MinorBlockChain.HasBlock(blockHash) {
 		return BLOCK_COMMITTED
 	}
-
-	//TODO support BLOCK_COMMITTING???
-
-	// Check if the block is being propagating to other slaves and the master
-	// Let's make sure all the shards and master got it before committing it
 	return BLOCK_UNCOMMITTED
 }
 
--- a/core/minorblockchain.go
+++ b/core/minorblockchain.go
@@ -537,8 +537,11 @@ func (m *MinorBlockChain) Genesis() *types.MinorBlock {
 }
 
 // HasBlock checks if a block is fully present in the database or not.
+// Both the commit marker and the block body must exist. A crash (OOM/kill)
+// can leave the marker present while the body is absent; returning true in
+// that case causes sync to skip re-downloading the block and stall.
 func (m *MinorBlockChain) HasBlock(hash common.Hash) bool {
-	return m.IsMinorBlockCommittedByHash(hash)
+	return m.IsMinorBlockCommittedByHash(hash) && rawdb.HasBlock(m.db, hash)
 }
 
 // HasState checks if state trie is fully present in the database or not.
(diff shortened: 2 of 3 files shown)


blockchaindevsh:
Verdict: approve the change; tighten one downstream caller before the fix is end-to-end useful.
                                                                                                                                                                                                                                                    
  Holds up                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                    
  - Root cause confirmed: IsMinorBlockCommittedByHash is just rawdb.HasCommitMinorBlock (core/minorblockchain_addon.go:1865-1867). Marker key "cmB"+hash and body key "b"+hash are at distinct prefixes (core/rawdb/schema.go:35,54), written in    
  separate non-batched Puts. Partial durability after OOM is expected LevelDB behavior.                                                                                                                                                           
  - Fix at core/minorblockchain.go:539-542 is minimal: marker check first (fast, more selective), body check on hit.                                                                                                                                
  - Quiet bonus: HasBlockAndState (:554-560) had a latent panic path — the old marker-only HasBlock returned true → GetMinorBlock(...) == nil → panic("bug fix block can not be nil") at :560. Now correctly short-circuits.                        
  - All callers tolerate the tightening (most are improved — sync re-download triggers, insertSidechain re-writes body, InsertReceiptChain no longer wrongly skips).                                                                                
  - Unit test TestHasBlockMarkerWithoutBodyReturnsFalse (core/minorblockchain_test.go:1366-1404) correctly purges blockCache (warm from InsertChain) and cross-checks rawdb.HasBlock == false && rawdb.HasCommitMinorBlock == true as a model       
  assertion.                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                    
  Pre-merge ask                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                  
  Tighten getBlockCommitStatusByHash in cluster/shard/shard.go:209-223. Verified post-PR: it still calls IsMinorBlockCommittedByHash only. Call path that breaks the docstring's promise:                                                           
  1. Sync's HasBlock correctly returns false (this PR) → block downloaded → handed to AddBlockListForSync.
  2. AddBlockListForSync at api_backend.go:448 calls getBlockCommitStatusByHash → returns BLOCK_COMMITTED (marker still there) → continue.                                                                                                          
  3. Body is never restored.                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                    
  One-liner fix: if s.MinorBlockChain.HasBlock(blockHash) { return BLOCK_COMMITTED }. May be in scope for one of #686/#687; confirm with author.                                                                                                    
                                                                                                                                                                                                                                                    
  Architectural follow-up                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                    
  - WriteBlockWithState writes body at :876 outside the batch at :959-980, then writes marker at :988 after batch.Write() returns. PR fixes the symptom; the underlying invariant break should be a follow-up (single batch carrying body + marker +
   receipts).
                                                                                                                                                                                                                                                    
  Nits                                                                                                                                                                                                                                            

  - Comment at :540 only describes one direction of the marker/body skew; the && is symmetric.                                                                                                                                                      
  - No test for the HasBlockAndState panic-elimination.
  - Future cache short-circuit (like RootBlockChain.HasBlock at core/rootblockchain.go:340-345) would silently regress this fix — a defensive subtest without Purge() would future-proof.                                                           
  - Rebase folds away the .github/workflows/rpc_healthcheck.yml noise.

syntrust (dismissed):

qzhodl (approved):

ping-ke:
Cover by PR https://github.com/QuarkChain/goquarkchain/pull/689. Close this one.

PR #683, opened 2026-06-09
Graceful shutdown for cluster scripts

ping-ke:
### Summary

- stop_cluster.sh previously used a single kill -9 which prevented the node from flushing RocksDB and closing connections cleanly. Replace with SIGTERM → wait → SIGKILL fallback.
- run_cluster.sh now writes each process PID to cluster.pids on startup (file is truncated on each run so stale entries are never inherited), enabling reliable PID-based shutdown.

### Changes

run_cluster.sh
- Create/truncate cluster.pids at startup
- Append each slave and master PID to the file after launch
- Remove eval-based command construction; run commands directly
- Fix indentation (mixed tabs/spaces → consistent 4-space)

stop_cluster.sh
- Rewrite from a one-liner to a full graceful-shutdown script:
  a. Read PIDs from cluster.pids; fall back to ps grep --cluster_config if the file is absent (manual start or old script)
  b. Send SIGTERM to all live processes
  c. Wait up to 60 s for clean exit, printing still-running PIDs every 10 s
  d. SIGKILL only processes that exceed the timeout
  e. Clean up cluster.pids on exit
- Add #!/bin/bash shebang and set executable bit on both scripts

### Test plan

  - [X] Start cluster with run_cluster.sh, verify cluster.pids is created with correct PIDs
  - [X] Run stop_cluster.sh — confirm processes exit cleanly (no SIGKILL log line under normal conditions)
  - [X] Delete cluster.pids manually, run stop_cluster.sh — confirm fallback ps path works


blockchaindevsh:
Verdict: approve with the websocket-sibling fix.
                                                                                                                                                                                                                                                    
  Holds up                                                                                                                                                                                                                                        

  - SIGTERM → 60s grace → SIGKILL fallback (stop_cluster.sh:30-67). Right shape.                                                                                                                                                                    
  - Go binary actually handles SIGTERM in cmd/utils/cmd.go:43 (stack.Stop() on SIGINT/SIGTERM); the courtesy is real.
  - PID-file primary + ps-grep fallback (:7-26), with empty-PID-file and "no processes" idempotency.                                                                                                                                                
  - run_cluster.sh:11-13 truncates PID file at start; :18-19,30 drops broken eval $cmd (latent quoting bug fix under spaces in configPath).                                                                                                         
  - Progress log [Ns] still waiting on pids: … every ~10s.                                                                                                                                                                                          
                                                                                                                                                                                                                                                    
  Pre-merge ask                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                  
  run_cluster_with_websocket.sh was not updated and still uses eval $cmd + writes no PID file. Verified via git diff --name-only. An operator who starts with the websocket variant and stops with the new script falls through to the looser       
  ps-grep path, partially defeating the contract. Either replicate the PID-file block in the websocket script or factor a shared helper.
                                                                                                                                                                                                                                                    
  Concerns                                                                                                                                                                                                                                        

  - TIMEOUT=60 hardcoded (stop_cluster.sh:4). A cluster mid-compaction can need more. TIMEOUT=${TIMEOUT:-60} is one line.                                                                                                                           
  - mapfile is Bash 4+ (:8,19). macOS ships Bash 3.2; verified the shebang is /bin/bash. Replace with portable while read loop (3 lines).
  - No non-zero exit on SIGKILL fallback (:59-67,72). systemd/ansible can't alarm on "shutdown wasn't graceful." Exit 1 in that branch.                                                                                                             
  - No process-group handling — only the immediate ./cluster PID is recorded; child processes that don't propagate signals would orphan. setsid + kill -TERM -$pgid is the belt-and-braces version.                                                 
  - ps-grep can false-positive on --cluster_config on shared hosts.                                                                                                                                                                                 
  - PID file is $PWD-dependent. Pin to script directory: PIDFILE="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)/cluster.pids".                                                                                                                      
                                                                                                                                                                                                                                                    
  Nits                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                    
  - .github/workflows/rpc_healthcheck.yml in diff is PR #679 rebase noise.                                                                                                                                                                          
  - Backticks \pwd`inrun_cluster.sh:16,27` — pre-existing.
  - Commit 6d0dd45 ("change chmod") squash into 8a7898b.                                                                                                                                                                                            
  - cluster.pids should be in .gitignore.

qzhodl:
Same as the above comments, two issues may need to be fixed before merge

**1. `run_cluster_with_websocket.sh` wasn't updated.** It's a near-identical twin that still uses `eval` and does not write to `cluster.pids`. So clusters started via the websocket script get no PID file, and `stop_cluster.sh` silently falls back to the `ps` path (losing the precise PID-based shutdown this PR adds, and inheriting the "stops all clusters on the machine" caveat). Please mirror the `PIDFILE` truncate + `echo $pid >> $PIDFILE` logic there too.

**2. `mapfile` requires bash ≥ 4 and fails silently on bash 3.2 (default macOS).** On 3.2, `mapfile` is `command not found`, leaving `pids` empty with exit 0 — so even when the PID file exists, `stop_cluster.sh` hits the `${#pids[@]} -eq 0` branch, prints "PID file is empty, nothing to stop.", deletes the PID file, and exits 0 without stopping anything. Fine on the Ubuntu deploy target, but a trap for local macOS dev. Either guard the version (`[ "${BASH_VERSINFO[0]}" -ge 4 ] || { echo "needs bash >= 4"; exit 1; }`) or use a portable read loop (`while IFS= read -r p; do pids+=("$p"); done < "$PIDFILE"`).

ping-ke:
> Same as the above comments, two issues may need to be fixed before merge
> 
> **1. `run_cluster_with_websocket.sh` wasn't updated.** It's a near-identical twin that still uses `eval` and does not write to `cluster.pids`. So clusters started via the websocket script get no PID file, and `stop_cluster.sh` silently falls back to the `ps` path (losing the precise PID-based shutdown this PR adds, and inheriting the "stops all clusters on the machine" caveat). Please mirror the `PIDFILE` truncate + `echo $pid >> $PIDFILE` logic there too.
> 
> **2. `mapfile` requires bash ≥ 4 and fails silently on bash 3.2 (default macOS).** On 3.2, `mapfile` is `command not found`, leaving `pids` empty with exit 0 — so even when the PID file exists, `stop_cluster.sh` hits the `${#pids[@]} -eq 0` branch, prints "PID file is empty, nothing to stop.", deletes the PID file, and exits 0 without stopping anything. Fine on the Ubuntu deploy target, but a trap for local macOS dev. Either guard the version (`[ "${BASH_VERSINFO[0]}" -ge 4 ] || { echo "needs bash >= 4"; exit 1; }`) or use a portable read loop (`while IFS= read -r p; do pids+=("$p"); done < "$PIDFILE"`).

Fixed.

syntrust:
Failed on macOS

```
grep: invalid option -- P

grep -V
grep (BSD grep, GNU compatible) 2.6.0-FreeBSD
```

qzhodl (dismissed):

blockchaindevsh:
> Failed on macOS
> 
> ```
> grep: invalid option -- P
> 
> grep -V
> grep (BSD grep, GNU compatible) 2.6.0-FreeBSD
> ```

@ping-ke shall we support macOS?

syntrust:
The scripts now work on macOS. May we need to update the README.md to include the stop script?

ping-ke:
> The scripts now work on macOS. May we need to update the README.md to include the stop script?

Added.

ping-ke:
> > Failed on macOS
> > ```
> > grep: invalid option -- P
> > 
> > grep -V
> > grep (BSD grep, GNU compatible) 2.6.0-FreeBSD
> > ```
> 
> @ping-ke shall we support macOS?

I think it would be better to support macOS, I have updated the script.

qzhodl (approved):

syntrust (approved):

PR #685, opened 2026-06-12
fix(serialize): reject oversized length prefixes to prevent OOM

ping-ke:
## Summary

A length field read from the wire was used to size memory allocations **before any bytes were validated**.

A malicious P2P message can declare a huge length and force the decoder to allocate tens of gigabytes before reading a single element, crashing the node via OOM or an unrecovered `makeslice` panic.

This PR validates every length prefix against the remaining buffer at the point it is read.

---

## Vulnerability Details

Inbound P2P/cluster payloads (`qkcMsg.Data`) are passed directly into `serialize.DeserializeFromBytes`.

Many command types — including the **pre-authentication** `HelloCmd` — contain slice fields whose length is encoded using a 4-byte length prefix:

```go
ChainMaskList []uint32 `bytesizeofslicelen:"4"`
```

In `deserializeList`:

```go
vlen, _ = bb.getLen(ts.ByteSizeOfSliceLen)        // attacker-controlled, up to ~4.29e9
newv := reflect.MakeSlice(val.Type(), vlen, vlen) // allocates BEFORE reading any element
```

Because the allocation happens before any element is decoded, an attacker can send a tiny message that declares an enormous list length and trigger massive memory allocation immediately.

For example, a ~40-byte packet with:

```text
vlen = 0xFFFFFFFF
```

can result in the following allocations:

| Field | Element Size | Eager Allocation |
|---------|---------|---------|
| `ChainMaskList []uint32` (`HelloCmd`) | 4 B | ~17 GB |
| `RootBlockHashList []common.Hash` | 32 B | ~137 GB |
| `[]*MinorBlockHeader` | 8 B | ~34 GB |

The existing per-element bounds checks never execute because `MakeSlice` occurs before the element-reading loop begins.

### Why Existing Limits Do Not Help

`DefaultP2PCmddSizeLimit` (128 MB) does not mitigate this issue.

The allocation size is derived from the **declared element count**, not the actual payload size. As a result, a very small, otherwise in-spec frame can trigger a huge allocation before any payload bytes are consumed.

---

## Additional Affected Code Path

`CrossShardTransactionDepositList.Deserialize` contains the same pattern:

```go
make([]*CrossShardTransactionDeposit, size)
```

The list length is obtained via:

```go
GetUInt32(...)
```

which bypasses `getLen()` entirely.

Although the size is capped at `MaxUint24` (~16.7 million entries), this can still trigger approximately:

```text
16.7M × 8 bytes ≈ 134 MB
```

of allocation before validating that sufficient bytes remain in the buffer.

---

## Fix

A length prefix can never legitimately exceed the number of bytes remaining in the buffer, since every list element (or byte slice entry) consumes at least one byte.

This PR adds validation that rejects any declared length larger than the remaining unread buffer **before allocation occurs**.

#### 1. `serialize/getLen`

Added a guard at the central length-decoding choke point:

- Protects all `deserializeList()` callers.
- Protects all `GetVarBytes()` callers.
- Prevents oversized `reflect.MakeSlice()` allocations.
- Ensures allocations remain bounded by the actual message size.

#### 2. `CrossShardTransactionDepositList.Deserialize`

Added the same validation before:

```go
make([]*CrossShardTransactionDeposit, size)
```

since this code path bypasses `getLen()`.

---

### New Tests

- `TestDeserializeRejectsOversizedSliceLength`
  - Uses a 4-byte length prefix of `0xFFFFFFFF`
  - No payload elements
  - Returns an error instead of triggering OOM/panic

- `TestDeserializeValidSliceStillWorks`
  - Verifies normal slice decoding remains unchanged

- `TestCrossShardTransactionDepositListRejectsOversizedLength`
  - Uses:
    - `version = 1`
    - `size = MaxUint24`
  - No payload elements
  - Returns an error

### Updated Tests

Updated 7 existing truncated-input test cases to expect the new, earlier validation error:

```text
length N exceeds remaining buffer M bytes
```

Fixed-size reads such as:

- `uint16`
- `Uint128`
- fixed-size arrays

remain unchanged and continue to be validated through `getBytes()`.

---

### Before

An unauthenticated peer could send a tiny packet containing a malicious length prefix and cause:

- Massive memory allocation
- OOM termination
- Unrecovered `makeslice` panic
- Remote denial-of-service

### After

Length prefixes are validated against the remaining unread buffer before allocation.

Memory allocation is therefore bounded by the actual message size, preventing oversized slice allocations and eliminating this denial-of-service vector.

### DIFF
--- a/core/types/cross_shard_transaction.go
+++ b/core/types/cross_shard_transaction.go
@@ -64,6 +64,12 @@ func (c *CrossShardTransactionDepositList) Deserialize(bb *serialize.ByteBuffer)
 	}
 	version := size >> 24
 	size = size & MaxUint24
+	// Guard against a malicious length prefix before allocating: each deposit
+	// serializes to more than one byte, so a count larger than the remaining
+	// buffer is invalid and could otherwise force a huge make()/OOM.
+	if int(size) > bb.Remaining() {
+		return fmt.Errorf("deser: deposit list length %d exceeds remaining buffer %d bytes", size, bb.Remaining())
+	}
 	txList := make([]*CrossShardTransactionDeposit, size)
 	switch version {
 	case 0:
--- a/core/types/cross_shard_transaction.go
+++ b/core/types/cross_shard_transaction.go
@@ -67,8 +67,13 @@ func (c *CrossShardTransactionDepositList) Deserialize(bb *serialize.ByteBuffer)
 	// Guard against a malicious length prefix before allocating: each deposit
 	// serializes to more than one byte, so a count larger than the remaining
 	// buffer is invalid and could otherwise force a huge make()/OOM.
-	if int(size) > bb.Remaining() {
-		return fmt.Errorf("deser: deposit list length %d exceeds remaining buffer %d bytes", size, bb.Remaining())
+	remaining := bb.Remaining()
+	if int(size) > remaining {
+		bytesStr := "bytes"
+		if remaining == 1 {
+			bytesStr = "byte"
+		}
+		return fmt.Errorf("deser: deposit list length %d exceeds remaining buffer %d %s", size, remaining, bytesStr)
 	}
 	txList := make([]*CrossShardTransactionDeposit, size)
 	switch version {
--- a/core/types/cross_shard_transaction_test.go
+++ b/core/types/cross_shard_transaction_test.go
@@ -132,3 +132,18 @@ func TestReadCrossShardTransactionDepositList(t *testing.T) {
 	}
 
 }
+
+// TestCrossShardTransactionDepositListRejectsOversizedLength is a regression
+// test for an OOM/DoS: the deposit-count prefix must be validated against the
+// remaining buffer before make() allocates the slice. The 4-byte prefix encodes
+// version=1 in the top byte and size=MaxUint24 (~16.7M) in the low 24 bits, with
+// no element bytes following. Without the guard this forces a ~134MB allocation
+// (16.7M * 8-byte pointers) before a single element is read.
+func TestCrossShardTransactionDepositListRejectsOversizedLength(t *testing.T) {
+	input := []byte{0x01, 0xFF, 0xFF, 0xFF}
+
+	d := NewCrossShardTransactionDepositList(nil)
+	err := serialize.DeserializeFromBytes(input, d)
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "exceeds remaining buffer")
+}
--- a/p2p/qkchandle.go
+++ b/p2p/qkchandle.go
@@ -9,6 +9,7 @@ import (
 	"encoding/binary"
 	"encoding/hex"
 	"errors"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"math/big"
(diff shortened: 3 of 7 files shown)


blockchaindevsh:
Verdict: approve; open two follow-ups before considering this surface "done."
                                                                                                                                                                                                                                                    
  Holds up
                                                                                                                                                                                                                                                    
  - Single choke-point check size > bb.Remaining() in ByteBuffer.getLen covers ~all length-prefixed reads via deserializeList, GetVarBytes, deserializeString, deserializeBigInt.                                                                   
  - One explicit guard in CrossShardTransactionDepositList.Deserialize (which bypasses getLen because the count and version byte share a uint32). Right placement.
  - Bound is reader-relative, not a magic absolute. Holds for every element type the codec accepts (1+ wire byte per element).                                                                                                                      
  - Backward compat preserved by construction: encoder writes [length][exactly-length-bytes], so for any well-formed payload length ≤ remaining.                                                                                                    
  - Returns error, not panic.                                                                                                                                                                                                                       
  - Tests target the precise attack (sliceLenVictim mirrors p2p.HelloCmd.ChainMaskList; 4-byte 0xFFFFFFFF payload).                                                                                                                                 
                                                                                                                                                                                                                                                    
  Pre-merge / immediate follow-ups                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                    
  1. Adjacent unfixed OOM at p2p/qkchandle.go:87 — verified: frameBuf := make([]byte, fSize) with fSize from the wire and no upper bound. MAC at :81-83 validates only the header bytes, not the size. Peer can claim 4 GiB. This is the easier     
  attack and remains open after this PR.
  2. getLen byteSize-8 overflow (latent) — verified at serialize/bytebuffer.go:70-94: loop is size = (size << 8) | int(b[i]) on Go int. For byteSize=8 on 64-bit, can overflow to negative; then size > Remaining() is negative > positive → false →
   guard bypass. No 8-byte callers today, but no upper bound on byteSize either. Either reject byteSize > 4 or compare in uint64.                                                                                                                   
  
  Smaller concerns                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                  
  - TokenBalances.Deserialize (core/types/token.go:337-359) reads num uint32 without pre-allocation but loops num times — indirectly safe (first iteration hits getLen) but symmetry-restoring guard would harden.                                  
  - Cap is loose for pointer-element slices ([]*Tx is ~8× amplification): 100 MB message can pin ~800 MB. Bounded, not exploitable.
  - Per-message attacker amplification still up to ~8× wire size — needs corroborating message-size cap (see #1) + per-peer rate limits.                                                                                                            
                                                                                                                                                                                                                                                    
  Nits                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                    
  - .github/workflows/rpc_healthcheck.yml rebase noise.                                                                                                                                                                                             
  - Error msg singular/plural bug ("1 bytes").                                                                                                                                                                                                    
  - Consider if vlen < 0 { return errInvalidLen } defensively in deserializeList.

syntrust (dismissed):

qzhodl (approved):

syntrust (approved):

PR #686, opened 2026-06-14
fix: resolve sync marker/body race and empty-xshard batch abandonment

ping-ke:
### Background

  Two distinct bugs in cluster/shard/api_backend.go are addressed.

  Bug 1 — marker/body race in AddMinorBlock / AddBlockListForSync

  The shard layer called CommitMinorBlockByHash outside m.chainmu, after InsertChainForDeposits had already returned. Concurrently, AddRootBlock → setHead (running under m.chainmu) could delete that same commit marker. The out-of-lock re-write could then land after the deletion, leaving the block in a marker=present, body=absent state. HasBlock() would return true for a block whose body was gone, and the next block would fail with unknown ancestor.

  Fix: remove the redundant shard-layer CommitMinorBlockByHash calls. The marker is now written exactly once, inside InsertChainForDeposits under m.chainmu, so setHead serializes cleanly against it.

  Bug 3 — batch abandoned on empty xshard list

  When InsertChainForDeposits returned an empty xshard list (len != 1), the old code did return nil, silently dropping every remaining block in the sync batch. An empty list has two distinct causes that must be handled differently:

  Fix: branch on getBlockCommitStatusByHash. Committed → continue; not committed → return an error naming the offending block.

###  Tests

  - core/minorblockchain_test.go
    - TestInsertChainForDepositsKnownBlockReturnsEmptyXshard — proves the real (empty xshard, nil error) return shape the shard-layer fix assumes is genuine, not invented.
  - cluster/shard/api_backend_test.go (new)
    - -race interleaving tests (*MarkerBodyConsistencyUnderRace, 100 iterations) that reproduce the Bug 1 race for both AddMinorBlock and AddBlockListForSync.
    - Seam-based tests (insertChainForDeposits indirection) driving both Bug 3 branches deterministically — *ContinuesAfterCommittedEmptyXshard (continue path) and *ErrorsOnNonCommittedEmptyXshard (error path) .

  A test-only seam, var insertChainForDeposits = (*core.MinorBlockChain).InsertChainForDeposits, is introduced so tests can simulate the sidechain-commit and procInterrupt outcomes. Production code never reassigns it.

  Test plan

  go test ./core/... 
  go test -race ./cluster/shard/... -run TestAddBlockListForSync
  go test -race ./cluster/shard/... -run TestAddMinorBlock

### DIFF
--- a/cluster/shard/api_backend.go
+++ b/cluster/shard/api_backend.go
@@ -363,7 +363,13 @@ func (s *ShardBackend) AddMinorBlock(block *types.MinorBlock) error {
 	}
 
 	if len(xshardLst) != 1 {
-		log.Warn(s.logInfo+" already have this block", "number", block.NumberU64(), "hash", block.Hash().String())
+		// Two cases produce an empty xshard list:
+		// 1. ErrPrunedAncestor → insertSidechain: block committed inside insertSidechain.
+		// 2. procInterrupt (chain shutting down): block not committed.
+		if s.getBlockCommitStatusByHash(block.Hash()) != BLOCK_COMMITTED {
+			return fmt.Errorf("InsertChainForDeposits returned empty xshard for non-committed block %d %s", block.NumberU64(), block.Hash().String())
+		}
+		log.Info(s.logInfo+" already have this block", "number", block.NumberU64(), "hash", block.Hash().String())
 		return nil
 	}
 
@@ -461,9 +467,9 @@ func (s *ShardBackend) AddBlockListForSync(blockLst []*types.MinorBlock) error {
 			// 2. procInterrupt (chain shutting down): block not committed.
 			//    Return error so the caller retries after the node is ready.
 			if s.getBlockCommitStatusByHash(blockHash) != BLOCK_COMMITTED {
-				return fmt.Errorf("InsertChainForDeposits returned empty xshard for non-committed block %d", block.NumberU64())
+				return fmt.Errorf("InsertChainForDeposits returned empty xshard for non-committed block %d %s", block.NumberU64(), blockHash.String())
 			}
-			log.Warn(s.logInfo+" already have this block", "number", block.NumberU64(), "hash", block.Hash().String())
+			log.Info(s.logInfo+" already have this block", "number", block.NumberU64(), "hash", block.Hash().String())
 			continue
 		}
 		s.mBPool.delBlockInPool(block.Hash())
--- a/cluster/shard/api_backend.go
+++ b/cluster/shard/api_backend.go
@@ -399,7 +399,6 @@ func (s *ShardBackend) AddMinorBlock(block *types.MinorBlock) error {
 		s.setHead(currHead.Number)
 		return err
 	}
-	s.MinorBlockChain.CommitMinorBlockByHash(block.Hash())
 	if s.MinorBlockChain.CurrentBlock().Hash() != currHead.Hash() {
 		go s.miner.HandleNewTip()
 	}
@@ -413,6 +412,13 @@ func (s *ShardBackend) AddMinorBlock(block *types.MinorBlock) error {
 
 	return nil
 }
+
+// insertChainForDeposits is a test seam over (*core.MinorBlockChain).InsertChainForDeposits.
+// AddBlockListForSync goes through this variable so tests can reproduce the two
+// empty-xshard outcomes (pruned-ancestor sidechain commit, and procInterrupt
+// shutdown) without constructing a 128-block pruned chain. Production code never
+// reassigns it.
+var insertChainForDeposits = (*core.MinorBlockChain).InsertChainForDeposits
 
 // Add blocks in batch to reduce RPCs. Will NOT broadcast to peers.
 //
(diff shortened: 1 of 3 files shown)


blockchaindevsh:
Verdict: approve; restore the deleted test before merge.
                                                                                                                                                                                                                                                    
  Holds up                                                                                                                                                                                                                                          
  
  - Race window genuinely closes. Post-PR the marker is written ONLY at core/minorblockchain.go:988 (WriteBlockWithState under m.mu + m.chainmu) or :1258 (insertSidechain under m.chainmu). Both serialize against setHead.                        
  - Removed outside-lock CommitMinorBlockByHash calls are truly redundant — every production block-insert path already writes the marker under lock.                                                                                              
  - Empty-xshard case correctly differentiated:                                                                                                                                                                                                     
    - procInterrupt (minorblockchain.go:1065-1067): nothing written → recheck BLOCK_UNCOMMITTED → error → retry.                                                                                                                                  
    - insertSidechain (:1270-1273): marker written, not promoted → recheck BLOCK_COMMITTED → continue.                                                                                                                                              
  - No "xshard gap" regression — silent no-broadcast in case-1 matches "not yet canonical"; when later promoted, the recursive insertChain at :1305/1319 bubbles the xshard list back up for broadcast. Mirrors AddMinorBlock's convention at     
  :365-368.                                                                                                                                                                                                                                         
  - Recheck race after chainmu release is benign (spurious "not committed" → caller retries; self-healing).                                                                                                                                       
                                                                                                                                                                                                                                                    
  Pre-merge ask                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                    
  Restore TestSetHeadDeletesBodyAndCommitMarker. Verified via git show 3405389: commit literally deletes the 40-line test from core/minorblockchain_test.go. That test asserted deterministically (no -race, no flake schedule) that                
  DeleteMinorBlock clears both body AND marker. This is the single most important guard against the Bug A invariant regressing. Without it, future changes that broke DeleteMinorBlock's atomicity would silently pass the seam-based shard tests
  (which stub the dependency) and the race tests would catch it only with -race AND a lucky schedule. Fix the original integration issue rather than deleting the assertion.                                                                        
                                                                                                                                                                                                                                                  
  Smaller concerns

  - Test seam global is fragile. var insertChainForDeposits = ... (api_backend.go:421) mutated with defer restore. Adding t.Parallel() later → data race. Wrap in a single capture/restore helper.                                                  
  - Error message at api_backend.go:464 missing the hash. Convention at :466 warn-log includes it.
  - AddMinorBlock vs AddBlockListForSync asymmetry. The new recheck guard at :463-465 wasn't mirrored to AddMinorBlock's equivalent at :365-368. Defensible (single-block path's procInterrupt drops the block and retries on next sync), but worth 
  flagging silence.                                                                                                                                                                                                                                 
  - log.Warn per-block on bulk sync at :466 can fire many times now that we continue instead of bail. Demote to Info/Debug.
                                                                                                                                                                                                                                                    
  Nits                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                    
  - Unicode → in comment, // ── stubs ── box-drawing — minor style drift.                                                                                                                                                                           
  - TestAddBlockListForSync_ContinuesPastKnownBlock (:179) honestly self-documents it covers a pre-existing branch, not Bug B. Decorative for this PR.
  - .github/workflows/rpc_healthcheck.yml rebase noise.

syntrust:
I think we should sort out the relationship between this PR and https://github.com/QuarkChain/goquarkchain/pull/687. They are not two independent fixes that can be merged in arbitrary order; they look more like adjacent fixes in the same execution path.

cluster/shard/api_backend.go
```diff
@@ -399,7 +405,6 @@ func (s *ShardBackend) AddMinorBlock(block *types.MinorBlock) error {
 		s.setHead(currHead.Number)
 		return err
 	}
-	s.MinorBlockChain.CommitMinorBlockByHash(block.Hash())
```
qzhodl, line 402: // If the block is committed, it means
	// - All neighbor shards/slaves receives x-shard tx list
	// - The block header is sent to master
	// then return immediately
According to the comments, we should not delete this code

qzhodl:
The race fix is real, but I think the **commit-marker semantics** mean this is fixing the symptom in the wrong layer. Cross-checking `pyquarkchain` (the reference):

In `shard.py::add_block`, `commit_by_hash` is called **exactly once, and only after propagation succeeds**:

```python
xshard_list, _ = self.state.add_block(block, force=True)      # ① local insert: writes BODY only
...
await self.slave.broadcast_xshard_tx_list(block, xshard_list, ...)   # ② xshard -> neighbors
await self.slave.send_minor_block_header_to_master(block.header, ...) # ③ header -> master
self.state.commit_by_hash(block_hash)                         # ④ commit ONLY after ②③
```

And the DB layer keeps two *separate* keys/concepts:
- `contain_minor_block_by_hash` → body present → "do we have this block" (used for ancestor checks)
- `is_minor_block_committed_by_hash` (`commit_` key) → "fully propagated"

`state.add_block` writes only the body; it never commits. So the marker genuinely means *"all neighbors got the xshard list AND master got the header"* — exactly what the comment copied into `getBlockCommitStatusByHash` says.

**goquarkchain has drifted from this in two places that happen to cancel out:**
1. `WriteBlockWithState` (core, ~minorblockchain.go:988) writes the commit marker **unconditionally at insert time** — the reference's `state.add_block` does not.
2. `MinorBlockChain.HasBlock` checks the **commit marker** (`IsMinorBlockCommittedByHash`), not the body — the reference checks the body (`contain_minor_block_by_hash`).

(1)+(2) make the marker effectively mean "inserted", which is why ancestor checks pass right after insert. The shard-layer `CommitMinorBlockByHash` this PR removes is therefore "redundant" *only under this drifted behavior*; under the reference design it is the **correct and only** commit point, and line 988 is the actual bug. (Note Go also dropped `BLOCK_COMMITTING` entirely — the propagation-window protection pyquarkchain relies on doesn't exist here.)

**Concrete risk this entrenches:** if a node crashes between ① insert and ② broadcast, the block is already marked committed (via 988), so on restart `AddMinorBlock` sees `BLOCK_COMMITTED` and early-returns → the xshard list is **never re-broadcast** → cross-shard deposits to neighbors can be lost. pyquarkchain avoids exactly this by committing only after ②③.

**Suggested direction** (bigger than this PR, but the correct shape): don't just delete the shard-layer commit —
1. remove the unconditional `CommitMinorBlockByHash` in `WriteBlockWithState` (and review the sidechain path at ~1258),
2. make `MinorBlockChain.HasBlock` check the body (`rawdb.HasBlock` / `GetMinorBlock != nil`), matching `contain_minor_block_by_hash`,
3. keep the shard-layer commit after broadcast+master (the correct point),
4. make that write race-safe vs `setHead` (hold `chainmu`, or re-check the block is still canonical before writing) — this is the legitimate core of Bug 1.

Just removing 988 alone won't work because of divergence (2): ancestor checks would then fail for a just-inserted-but-not-yet-propagated block. The two concepts pyquarkchain keeps separate (have-block vs committed) need to be split back apart.

The Bug 3 (empty-xshard batch continue) part of this PR looks correct and independent of the above.

PR #687, opened 2026-06-17
fix: replace setHead with RollbackToBlock and fix same-chain reorg

ping-ke:
###  Summary

  The post-insertion failure recovery in ShardBackend (xshard broadcast / master-notify / stats failures) had two problems:

  1. `AddBlockListForSync` never rolled back on failure. When `BatchBroadcastXshardTxList` or `SendMinorBlockHeaderListToMaster` failed, the code just `return err` — the blocks stayed inserted and canonical but uncommitted (`CommitMinorBlockByHash` only runs on success). On retry, `InsertChainForDeposits` hits `ErrKnownBlock` → `len(xshardLst) != 1` → early return `nil`, so `SendMinorBlockHeaderListToMaster` is skipped and the master never learns the block's coinbase — a permanent deadlock.
  2. `AddMinorBlock` recovered via `setHead`, which is heavier and riskier than needed. `setHead` calls `DeleteMinorBlock` (deletes body, receipts, and commit marker) plus `DeleteCanonicalHash` for every block above the target. That forces full state re-execution on retry, and it carries a "rewound state missing → reset to genesis" branch that can silently strand the chain at height 0.

  This branch introduces `RollbackToBlock`, which un-commits displaced blocks just enough to make them cleanly re-insertable (keeping body/receipts), wires it into both recovery paths, and fixes two persistence-ordering bugs in the same-chain reorg path.

###  Changes

  1. New `MinorBlockChain.RollbackToBlock(targetHash)` (`core/minorblockchain.go`)
  - For each block above the target, deletes only the commit marker (`DeleteMinorBlockCommitStatus`) and canonical-hash mapping (`DeleteCanonicalHash`) — keeps block body and receipts, so re-insertion skips re-writing bodies.
  - Verifies `targetHash` is an ancestor of the current head before rolling back.
  - Persists the new head via `WriteHeadBlockHash(targetHash)` at the end, so a crash after rollback restores the correct head rather than the block that failed broadcast.
  - Does not acquire `chainmu`; the caller must hold the shard mu lock.

  Contrast with `setHead`, which deletes bodies/receipts (forcing re-execution) and has the genesis-reset footgun.

  2. ShardBackend recovery uses RollbackToBlock (cluster/shard/api_backend.go)
  - `AddBlockListForSync`: records prevHead before the insertion loop and rolls back on `BatchBroadcastXshardTxList` / `SendMinorBlockHeaderListToMaster` failure — closing the master-coinbase deadlock.
  - `AddMinorBlock`: replaces the three `setHead(currHead.Number)` recovery calls (`BroadcastXshardTxList`, `GetShardStats`, `SendMinorBlockHeaderToMaster`) with `RollbackToBlock(currHead.Hash())`.
  - `broadcastNewTip` failure is demoted to a warn with no rollback — the block is fully committed locally and peers sync later; rolling back here would discard valid
  work.
  - Removed the now-unused `setHead` helper (which panic'd on error).

  3. Fixed same-chain reorg persistence (`core/minorblockchain.go`, `reorg()`)
  - The old code added `DeleteCanonicalHash` writes to the batch after `batch.Write()` had already `Destroy()`'d the underlying RocksDB WriteBatch, so the displaced canonical hashes were never persisted. Deletions now happen before `batch.Write()`.
  - `m.insert()` for the new head is now called after `batch.Write()`, so the head pointer is flushed only after the canonical-hash deletes and tx-index removals — matching the cross-chain path's ordering.

  4. `insert()` writes canonical + head hash atomically (`core/minorblockchain.go`) via a single batch, with error logging instead of silent failure.

  5. `setHead` genesis-reset removed (`core/minorblockchain.go`): the "rewound state missing → reset to genesis" branch is dropped. `InitFromRootBlock` rebuilds missing state via `reRunBlockWithState` before calling `setHead`; resetting to genesis here would undo that and strand the chain at height 0. The missing-state case is now the
  caller's responsibility.

  6. Minor fix (`core/minorblockchain_addon.go`): `reRunBlockWithState`'s "missing block" error no longer dereferences the nil block; it reports the parent hash instead.

###  Tests (`core/minorblockchain_test.go`)

  - `TestRollbackToBlock`: asserts commit markers + canonical hashes are cleared, bodies are kept, `currentBlock` is restored to target, and re-insertion succeeds.
  - `TestSameChainReorgClearsOldCanonicalHashes`: drives `reorg(block5,` block3) directly (an already-canonical `InsertChain` would hit the `ErrKnownBlock` path and never call `reorg`) and asserts displaced canonical hashes at heights 4 & 5 are cleared.

### DIFF
--- a/core/minorblockchain.go
+++ b/core/minorblockchain.go
@@ -730,33 +730,38 @@ func (m *MinorBlockChain) procFutureBlocks() {
 }
 
 // RollbackToBlock rolls back the chain so that the block with targetHash
-// becomes the new current head. All blocks above targetHash (up to the
-// current tip) are removed from the database — body, receipts, commit
-// marker, and canonical hash mapping — so that HasBlock returns false
-// for each removed block and they can be re-inserted later.
+// becomes the new current head. Blocks above targetHash have their commit
+// marker and canonical hash mapping removed so they can be re-inserted later.
+// Block body and receipts are kept to avoid re-executing state.
 //
-// Unlike SetHead, this method does not acquire chainmu or purge caches.
-// It is intended for fast rollback in the shard layer (e.g. when
-// broadcasting xshard txs fails after a block was inserted). The caller
-// must hold the shard's mu lock to prevent concurrent insert/setHead.
+// The caller must ensure targetHash is an ancestor of the current head
+// (same chain). RollbackToBlock does not acquire chainmu; the caller must
+// hold the shard's mu lock to prevent concurrent inserts.
 func (m *MinorBlockChain) RollbackToBlock(targetHash common.Hash) error {
 	target := m.GetMinorBlock(targetHash)
 	if target == nil {
 		return fmt.Errorf("RollbackToBlock: target block %x not found", targetHash)
 	}
 
-	for cur := m.CurrentBlock(); cur != nil && cur.Hash() != targetHash; cur = m.CurrentBlock() {
-		if cur.NumberU64() < target.NumberU64() {
-			return fmt.Errorf("RollbackToBlock: overshot target %x (cur %d < target %d)", targetHash, cur.NumberU64(), target.NumberU64())
+	// Verify target is on the same chain (ancestor of current head).
+	for cur := m.CurrentBlock(); ; cur = m.GetMinorBlock(cur.ParentHash()) {
+		if cur == nil {
+			return fmt.Errorf("RollbackToBlock: target %x is not an ancestor of current head", targetHash)
 		}
+		if cur.Hash() == targetHash {
+			break
+		}
+		if cur.NumberU64() < target.NumberU64() || cur.NumberU64() == 0 {
+			return fmt.Errorf("RollbackToBlock: target %x is not an ancestor of current head", targetHash)
+		}
+	}
+
+	for cur := m.CurrentBlock(); cur != nil && cur.Hash() != targetHash; cur = m.CurrentBlock() {
 		parent := m.GetMinorBlock(cur.ParentHash())
 		if parent == nil {
 			return fmt.Errorf("RollbackToBlock: parent missing for block %d [%x]", cur.NumberU64(), cur.Hash())
 		}
-
 		batch := m.db.NewBatch()
-		rawdb.DeleteMinorBlock(batch, cur.Hash())
-		rawdb.DeleteReceipts(batch, cur.Hash())
 		rawdb.DeleteMinorBlockCommitStatus(batch, cur.Hash())
 		rawdb.DeleteCanonicalHash(batch, rawdb.ChainTypeMinor, cur.NumberU64())
 		if err := batch.Write(); err != nil {
@@ -764,6 +769,7 @@ func (m *MinorBlockChain) RollbackToBlock(targetHash common.Hash) error {
 		}
 		m.currentBlock.Store(parent)
 	}
+	rawdb.WriteHeadBlockHash(m.db, targetHash)
 	return nil
 }
 
@@ -1473,11 +1479,15 @@ func (m *MinorBlockChain) reorg(oldBlock, newBlock types.IBlock) error {
 		for i := len(oldChain) - 1; i >= 0; i-- {
 			rawdb.DeleteCanonicalHash(batch, rawdb.ChainTypeMinor, oldChain[i].NumberU64())
 		}
-		m.insert(newBlock.(*types.MinorBlock))
-		log.Warn(m.logInfo+" same chain reorg, set currentBlock", "number", newBlockNumber, "hash", newBlock.Hash())
 	}
 
 	batch.Write()
+
+	if len(newChain) == 0 {
+		// insert after batch.Write() so the head pointer is persisted after canonical deletes
+		m.insert(newBlock.(*types.MinorBlock))
+		log.Warn(m.logInfo+" same chain reorg, set currentBlock", "number", newBlockNumber, "hash", newBlock.Hash())
+	}
 
 	// Insert the new chain, taking care of the proper incremental order
 	for i := len(newChain) - 1; i >= 0; i-- {
--- a/core/minorblockchain.go
+++ b/core/minorblockchain.go
@@ -320,19 +320,19 @@ func (m *MinorBlockChain) setHead(head uint64) error {
 	// Rewind the header chain, deleting all block bodies until then
 	batch := m.db.NewBatch()
 	for block := m.CurrentBlock(); block != nil && block.NumberU64() > head; block = m.CurrentBlock() {
+		preBlock := m.GetMinorBlock(block.ParentHash())
+		if preBlock == nil {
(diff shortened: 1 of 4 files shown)


core/minorblockchain.go
```diff
@@ -725,6 +729,50 @@ func (m *MinorBlockChain) procFutureBlocks() {
...
+		}
+	}
+
+	for cur := m.CurrentBlock(); cur != nil && cur.Hash() != targetHash; cur = m.CurrentBlock() {
+		parent := m.GetMinorBlock(cur.ParentHash())
+		if parent == nil {
+			return fmt.Errorf("RollbackToBlock: parent missing for block %d [%x]", cur.NumberU64(), cur.Hash())
+		}
+		batch := m.db.NewBatch()
+		rawdb.DeleteMinorBlockCommitStatus(batch, cur.Hash())
```
syntrust, line 763: May we need `removeTxIndexFromBlock " when deleting the minor block here.

core/minorblockchain.go
```diff
@@ -725,6 +729,50 @@ func (m *MinorBlockChain) procFutureBlocks() {
 	}
 }
 
+// RollbackToBlock rolls back the chain so that the block with targetHash
+// becomes the new current head. Blocks above targetHash have their commit
+// marker and canonical hash mapping removed so they can be re-inserted later.
+// Block body and receipts are kept to avoid re-executing state.
+//
+// The caller must ensure targetHash is an ancestor of the current head
+// (same chain). RollbackToBlock does not acquire chainmu; the caller must
```
syntrust: Semantically speaking, RollbackToBlock should take `chainmu` because it modifies the chain head/canonical/commit marker, which is the same type of chain structure write operation as SetHead and InsertChain.  
The distinction between `chainmu` and `mu` has become blurred, causing maintenance burdens.

blockchaindevsh:
Verdict: conditional accept; address the four pre-merge items.
                                                                                                                                                                                                                                                    
  Holds up                                                                                                                                                                                                                                        

  - Same-chain reorg bug is real, via AddRootBlock not InsertChain. _addon.go:1099-1134 → reWriteBlockIndexTo(N, M) → m.reorg(N, M) hits the len(newChain) == 0 branch. Master's setHead(M) did persist the canonical deletion correctly — with the 
  now-removed aggressive parts (body deletion + genesis reset).
  - AddBlockListForSync deadlock close is the most important fix. Master leaves all inserted blocks committed after broadcast failure → next retry continues every block → empty blockHashToXShardList → master never learns about the headers.     
  Permanent state divergence. RollbackToBlock(prevHead.Hash()) on failure restores re-insertability.                                                                                                                                                
  - Demoting broadcastNewTip failure to a warn (api_backend.go:412-415) is sound — by then SendMinorBlockHeaderToMaster succeeded; gossip can be retried.
  - Second commit's WriteHeadBlockHash (minorblockchain.go:772) and ordering fix (m.insert after batch.Write at :1484-1490) are necessary durability fixes.                                                                                         
  - Removes the setHead → genesis-reset on missing state correctly.                                                                                                                                                                                 
  - TestRollbackToBlock and TestSameChainReorgClearsOldCanonicalHashes cover previously-untested paths.                                                                                                                                             
                                                                                                                                                                                                                                                    
  Pre-merge punch list                                                                                                                                                                                                                            

  1. Body/receipts preservation is dead-key bloat, not the benefit the comment claims. Verified: RollbackToBlock at minorblockchain.go:765 deletes commit marker + canonical hash but NOT body/receipts. Doc claims "avoids re-executing state."    
  Doesn't survive inspection:
  - Re-insertion overwrites bodies unconditionally (WriteMinorBlock at :876, :832).                                                                                                                                                                 
  - Validator short-circuit uses commit marker (HasBlockAndState), not body — once marker is cleared, re-insertion proceeds regardless of body presence.                                                                                            
  - reRunBlockWithState walks backward from currentBlock; bodies above are never read.                                                                  
                                                                                                                                                                                                                                                    
  Verified via git show: first commit 1838924 did call DeleteMinorBlock (correct); second commit 656f961 removed it. Restore the first commit's behavior, or rewrite the doc to admit dead-row cost.                                                
                                                                                                                                                                                                                                                    
  2. insert() error handling regression (minorblockchain.go:526-536). Master's two unbatched calls each used log.Crit (process-terminating). PR batches them and log.Errors the failure — then m.currentBlock.Store(block) runs anyway. In-memory   
  tip advances while disk head pointer doesn't; on restart loadLastState silently rolls the chain back after the node has been serving the new tip. Restore log.Crit, or don't advance currentBlock on batch.Write failure and propagate the error. 
                                                                                                                                                                                                                                                    
  3. Locking semantics: s.mu doesn't block procFutureBlocks. Verified at minorblockchain.go:712-726: procFutureBlocks iterates futureBlocks LRU and calls InsertChain directly — does NOT take s.mu. Master's SetHead acquired m.chainmu at         
  :312-314, so this race didn't exist. The PR introduces it: RollbackToBlock clears canonical hashes / commit markers while a concurrent InsertChain from procFutureBlocks could be writing them. Acquire m.chainmu inside RollbackToBlock (matches
  SetHead's contract); the doc would then be accurate too.                                                                                                                                                                                          
                                                                                                                                                                                                                                                  
  4. Misleading commit message on 1838924. Asserts "DeleteCanonicalHash writes were added to a batch after batch.Write() had already called Destroy() on the underlying RocksDB WriteBatch." Master's reorg did not write after Destroy — the       
  deletions happened in the separately-batched setHead call. The actual fix is "compensating for removing the in-line setHead with inline canonical-hash deletes." Rewrite so future archaeologists don't waste time looking for an after-Destroy
  bug that wasn't there.                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                  
  Smaller concerns

  - RollbackToBlock writes one batch per block (N round-trips for N blocks; partial-failure leaves an unrecoverable intermediate state). Collect all deletions into one batch.
  - batch.Write() errors silently ignored in reorg at :1484. Pre-existing pattern, but now carries critical canonical-deletion data.
  - setHead/SetHead now dead from the same-chain-reorg path but still kept. Only remaining caller is ResetWithGenesisBlock (always head=0). Could be simplified to ResetToGenesis.                                                                  
                                                                                                                                                                                                                                                    
  Nits                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                    
  - Three near-identical rollback blocks in api_backend.go:380-411 — extract a rollbackOrLog(target, ctx) helper.                                                                                                                                   
  - Error wraps lose block context.
  - block4 declared but unused in TestSameChainReorgClearsOldCanonicalHashes.                                                                                                                                                                       
  - reRunBlockWithState change at _addon.go:539-545 is a quiet incidental nil-deref fix the commit doesn't mention.                                                                                                                                 
  - gofmt-style comment whitespace changes in test file inflate the diff.                                                                                                                                                                           
  - .github/workflows/rpc_healthcheck.yml rebase noise (PR #684).

PR #689, opened 2026-06-26
fix(commit-status): overhaul minor-block commit-status pipeline

ping-ke:
### Summary

Fixes the minor-block commit-status pipeline so a block is only treated as committed after its body exists and shard-layer coordination has completed. This prevents orphaned commit markers, lets sync retry partial failures, and avoids dropping later blocks in a sync batch.

### Changes

- Make `HasBlock` mean body presence again (`rawdb.HasBlock`), separate from the commit marker.
- Make `HasCommittedBlock` require both body presence and commit marker.
- Move commit-marker writes out of core import paths and into shard-layer success paths:
  - `AddMinorBlock` after x-shard broadcast + master header submission
  - `AddBlockListForSync` after batch x-shard broadcast + master header-list submission
  - genesis init remains a direct marker write
- Make `CommitMinorBlockByHash` locked and body-aware; it now returns `false` if the body was deleted before the marker could be written.
- Fix sync batch handling:
  - already-committed blocks are skipped instead of aborting the batch
  - duplicate blocks in one batch are ignored
  - body-present / marker-missing retries can re-execute read-only to recover x-shard lists
- Allow minor sync ancestor search to use `HasCommittedBlock || HasBodyWithoutState`, while skip logic still requires committed markers.
- Handle live `NewMinorBlock` when the parent body+state exists but the marker is missing by committing contiguous uncommitted ancestors first.
- Tighten `AddRootBlock` / `CreateBlockToMine` locking and snapshots to avoid mixed root-tip/current-block observations.

### Tests

Added coverage for:

- sync continuing past already-committed blocks
- commit markers written after `AddMinorBlock` / `AddBlockListForSync`
- body/marker consistency under concurrent rewind
- retry recovery when body exists but marker is missing
- `HasBlock` returning false for marker-only records
- `CommitMinorBlockByHash` returning false when the body is absent
- minor ancestor behavior for body-only/pruned-state anchors

### Notes

`CommitMinorBlockByHash` now returns `bool`; callers must handle `false` as “body missing, commit not written”.


core/minorblockchain_addon.go
```diff
@@ -1870,8 +1886,21 @@ func (m *MinorBlockChain) IsMinorBlockCommittedByHash(h common.Hash) bool {
...
-func (m *MinorBlockChain) CommitMinorBlockByHash(h common.Hash) {
+// CommitMinorBlockByHash writes the commit marker only when the block body is
+// still present in the DB. The check-and-write is performed under m.mu so it
+// is atomic with any concurrent operation that also holds m.mu while modifying
+// the chain (e.g. rewinding). Returns false when the body is absent (concurrent
+// reorg deleted it); the caller should treat the block as UNCOMMITTED and retry.
+// Callers that already hold m.mu (e.g. InitGenesisState) must call
+// rawdb.WriteCommitMinorBlock directly to avoid re-entrant locking.
+func (m *MinorBlockChain) CommitMinorBlockByHash(h common.Hash) bool {
+	m.mu.Lock()
```
qzhodl, line 1924: m.mu does not make this check-and-write atomic with SetHead: SetHead only holds chainmu while deleting the body and marker. It can delete the body between HasBlock and WriteCommitMinorBlock, recreating marker=yes, body=no. Please synchronize both paths with the same lock.

cluster/shard/api_backend.go
```diff
@@ -434,43 +462,96 @@ func (s *ShardBackend) AddBlockListForSync(blockLst []*types.MinorBlock) error {
...
 		}
 		if len(xshardLst) != 1 {
-			log.Warn(s.logInfo+" already have this block", "number", block.NumberU64(), "hash", block.Hash().String())
-			return nil
+			log.Warn(s.logInfo+" block already in DB, skipping", "number", block.NumberU64(), "hash", block.Hash().String())
+			// InsertChain skipped re-execution because the block body already exists.
+			// The outgoing xShard list cannot be recovered without re-executing the
+			// block (pyquarkchain uses force=True to handle this; goquarkchain does not
+			// yet support forced re-execution). Drop this block from the batch; it will
+			// be re-synced in a subsequent round.
```
qzhodl: After InsertChainForDeposits writes the body, a failure in BatchBroadcastXshardTxList or SendMinorBlockHeaderListToMaster leaves body=yes, marker=no. Neither outer entry point can retry it: SlaveBackend.AddBlockListForSync filters the hash because HasBlock now returns true for body presence, and ShardBackend.NewMinorBlock returns early for the same reason. Even a direct AddBlockListForSync retry treats it as a known block and cannot rebuild xshardLst. Please base these guards on commit status and support forced re-execution or persisted x-shard replay.

cluster/shard/api_backend.go
```diff
@@ -346,14 +346,34 @@ func (s *ShardBackend) AddMinorBlock(block *types.MinorBlock) error {
...
+	blockHash := block.Hash()
+
+	// Fast path (lock-free): check if already committed or being processed
+	commitStatus := s.getBlockCommitStatusByHash(blockHash)
+	if commitStatus == BLOCK_COMMITTED {
+		return nil
+	}
+	if commitStatus == BLOCK_COMMITTING {
+		log.Debug(s.logInfo+" Block is being processed by another goroutine, skipping", "hash", blockHash.Hex())
+		return nil // Skip processing, let the other goroutine handle it
```
qzhodl: Returning nil acknowledges success before the in-flight operation finishes. If that operation later fails, miner/sync callers do not receive the failure and this duplicate cannot take over after s.mu is released. Please wait for a shared result or let the caller block on s.mu instead of reporting success.

qzhodl (changes requested):
Please address the three blocking commit-status issues noted inline.

cluster/shard/api_backend.go
```diff
@@ -346,14 +346,34 @@ func (s *ShardBackend) AddMinorBlock(block *types.MinorBlock) error {
...
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	s.wg.Add(1)
 	defer s.wg.Done()
-	if commitStatus := s.getBlockCommitStatusByHash(block.Hash()); commitStatus == BLOCK_COMMITTED {
+
+	// Double-check after acquiring lock to avoid TOCTOU race
+	commitStatus = s.getBlockCommitStatusByHash(blockHash)
+	if commitStatus == BLOCK_COMMITTED || commitStatus == BLOCK_COMMITTING {
 		return nil
```
blockchaindevsh, line 401: Here also needs to handle the early return case mentioned above for `BLOCK_COMMITTING`.

core/minorblockchain.go
```diff
@@ -1259,7 +1260,7 @@ func (m *MinorBlockChain) insertSidechain(it *insertIterator, isCheckDB bool) (i
 			if err := m.WriteBlockWithoutState(block); err != nil {
 				return it.index, nil, nil, nil, err
 			}
-			m.CommitMinorBlockByHash(block.Hash())
+			// Note: CommitMinorBlockByHash is now only called at shard layer
```
blockchaindevsh: It's also called in `InitGenesisState` below.

cluster/shard/api_backend.go
```diff
@@ -434,43 +462,96 @@ func (s *ShardBackend) AddBlockListForSync(blockLst []*types.MinorBlock) error {
...
-			log.Warn(s.logInfo+" already have this block", "number", block.NumberU64(), "hash", block.Hash().String())
-			return nil
+			log.Warn(s.logInfo+" block already in DB, skipping", "number", block.NumberU64(), "hash", block.Hash().String())
+			// InsertChain skipped re-execution because the block body already exists.
+			// The outgoing xShard list cannot be recovered without re-executing the
+			// block (pyquarkchain uses force=True to handle this; goquarkchain does not
+			// yet support forced re-execution). Drop this block from the batch; it will
+			// be re-synced in a subsequent round.
+			// TODO: add force re-execution to InsertChainForDeposits to align with pyquarkchain.
+			s.processingBlocks.Delete(blockHash)
```
blockchaindevsh: Here should also clear the mark for uncommittedBlockHeaderList, maybe define a clear func then re-use for the two errors within this loop?

core/minorblockchain.go
ping-ke: Remove this `Note`.

core/minorblockchain.go
```diff
@@ -1158,7 +1157,8 @@ func (m *MinorBlockChain) insertChain(chain []types.IBlock, verifySeals bool, is
 			return it.index, events, coalescedLogs, xShardList, err
 		}
 
-		if isCheckDB {
+		if force && m.HasBlock(mBlock.Hash()) {
+			// Body already exists — re-execution only, do not write state again.
 			xShardList = append(xShardList, state.GetXShardList())
 			return 0, events, coalescedLogs, xShardList, nil
```
qzhodl: force=true still accepts a multi-block chain, but this return exits insertChain after the first block whose body already exists, silently dropping every remaining block while reporting success. Please continue after appending the recovered x-shard list, or explicitly reject force=true when len(chain) > 1. A regression test with [existingBlock, newBlock] would catch this.

core/minorblockchain_addon.go
```diff
@@ -1870,8 +1886,20 @@ func (m *MinorBlockChain) IsMinorBlockCommittedByHash(h common.Hash) bool {
...
 
-func (m *MinorBlockChain) CommitMinorBlockByHash(h common.Hash) {
+// CommitMinorBlockByHash writes the commit marker only when the block body is
+// still present in the DB. The check-and-write is performed under m.mu.
+// Returns false when the body is absent; the caller should treat the block
+// as UNCOMMITTED and retry.
+// Callers that already hold m.mu must call rawdb.WriteCommitMinorBlock
+// directly to avoid re-entrant locking.
+func (m *MinorBlockChain) CommitMinorBlockByHash(h common.Hash) bool {
+	m.mu.Lock()
```
qzhodl, line 1924: SetHead uses chainmu, so m.mu does not make this check-and-write atomic with body deletion. It can delete the body after HasBlock and before WriteCommitMinorBlock, leaving marker=yes/body=no. Please hold chainmu.RLock() before m.mu for this whole operation.

cluster/shard/api_backend.go
```diff
@@ -346,17 +346,28 @@ func (s *ShardBackend) AddMinorBlock(block *types.MinorBlock) error {
...
+	// Double-check after acquiring lock.
+	if s.getBlockCommitStatusByHash(blockHash) == BLOCK_COMMITTED {
 		return nil
 	}
-	//TODO support BLOCK_COMMITTING
+
 	currHead := s.MinorBlockChain.CurrentBlock().Header()
 	log.Debug(s.logInfo, "currHead", currHead.Number)
-	_, xshardLst, err := s.MinorBlockChain.InsertChainForDeposits([]types.IBlock{block}, false)
+	_, xshardLst, err := s.MinorBlockChain.InsertChainForDeposits([]types.IBlock{block}, true)
```
qzhodl: body=yes, marker=no is still blocked before this forced recovery path at both normal entry points: ShardBackend.NewMinorBlock returns at cluster/shard/api_backend.go:299-302 on HasBlock, and minor tip sync returns/skips at cluster/sync/minor_task.go:48-50 and 106-109 for the same reason. Only the root-block sync filter was fixed. Please make both entry points use commit status so peer rebroadcast and tip sync can retry an uncommitted body.

core/minorblockchain_addon.go
```diff
@@ -1361,8 +1377,8 @@ func (m *MinorBlockChain) getBlockCountByHeight(height uint64) int {
 
 // reWriteBlockIndexTo : already locked
 func (m *MinorBlockChain) reWriteBlockIndexTo(oldBlock *types.MinorBlock, newBlock *types.MinorBlock) error {
-	m.chainmu.Lock()
-	defer m.chainmu.Unlock()
+	m.mu.Lock()
```
syntrust, line 1398: The `m.mu` lock is no longer serialized against chainmu-based `setHead`.

If concurrently,
- goroutine A in `AddRootBlock` → `reWriteBlockIndexTo`→`reorg`→`setHead/insert` under `m.mu`;
- goroutine B in `AddMinorBlock` rollback → `setHead` under `chainmu`. 

Two writers to `currentBlock` + the canonical index under different locks → data race / canonical-chain corruption.

ping-ke: Change SetHead to `m.mu` lock, which consistent with geth 1.8.20

ping-ke:
# MinorBlockChain Lock Call Chains

> Verified against commit 6667712. Line numbers drift as the code changes.

## Locks

| Lock | Owner | Purpose | Declaration |
|------|-------|---------|-------------|
| `m.mu` | `MinorBlockChain` | Global chain-state lock (RWMutex) | `minorblockchain.go:109` |
| `m.chainmu` | `MinorBlockChain` | Block-insertion lock (RWMutex) | `minorblockchain.go:110` |
| `s.mu` | `ShardBackend` | Shard-level serialization lock (Mutex) | `shard.go:59` |

---

## 1. GetUnconfirmedHeaderList

```
cluster/slave/api_backend.go  SlaveBackend.GetUnconfirmedHeaderList
|-> cluster/shard/api_backend.go:281  ShardBackend.GetUnconfirmedHeaderList
      |-> core/minorblockchain_addon.go:783  MinorBlockChain.GetUnconfirmedHeaderList
            (m.mu.Lock)
```

---

## 2. AddRootBlock

```
cluster/slave/api_backend.go  SlaveBackend.AddRootBlock
|-> cluster/shard/api_backend.go:203  ShardBackend.AddRootBlock
      (s.mu.Lock)                                              ← serializes root/minor handling for the same shard
      |-> core/minorblockchain_addon.go:992  MinorBlockChain.AddRootBlock
            |-> 1. pre-checks + putRootBlock                    [no lock]
            |-> 2. no-change-to-root-tip early return :1079
            |         (m.mu.Lock snapshot rootTip; m.mu.Unlock)
            |-> 3. m.chainmu.Lock :1098 (deferred to end of fn) ← mutually exclusive with the insertChain pipeline
            |-> 4. m.mu.Lock :1103
            |         update rootTip / confirmedHeaderTip
            |         rewind currentBlock onto the same root chain
            |         compute needGenesisReset :1135
            |      m.mu.Unlock :1138
            |-> 5. [needGenesisReset] m.Reset :1158
            |         |-> ResetWithGenesisBlock
            |               |-> setHead(0)                      [unlocked; holds chainmu only]
            |-> 6. reWriteBlockIndexTo :1176
                      (m.mu.Lock — reorg + publish currentEvmState)
```

> Key: steps 5 and 6 run **after** `m.mu.Unlock` (1138).
> `Reset` uses the **unlocked `setHead`** (does not re-acquire chainmu/mu); `reWriteBlockIndexTo` takes only mu.
> chainmu is held throughout, so the reorg is mutually exclusive with the insertChain pipeline; no re-entrancy, no self-deadlock.

---

## 3. CreateShards → initGenesisState

```
cluster/slave/api_backend.go  SlaveBackend.CreateShards
|-> cluster/shard/api_backend.go:191  ShardBackend.InitFromRootBlock
      |-> cluster/shard/shard.go:180  ShardBackend.initGenesisState
            |-> 1. core/minorblockchain_addon.go:329  MinorBlockChain.InitGenesisState
            |         (m.mu.Lock)
            |-> 2. conn.BroadcastXshardTxList                   [no lock]
            |-> 3. core/minorblockchain_addon.go:1214  GetShardStats
            |         |-> getBlockCountByHeight :1379
            |                   (m.mu.RLock)
            |-> 4. conn.SendMinorBlockHeaderToMaster             [no lock]
```

---

## 4. NewMinorBlock → AddMinorBlock

```
cluster/slave/api_backend.go  SlaveBackend.NewMinorBlock
|-> cluster/shard/api_backend.go:303  ShardBackend.NewMinorBlock
      |   (HasCommittedBlock precheck; no lock)
      |-> cluster/shard/api_backend.go:365  ShardBackend.AddMinorBlock
            |-> 0. fast-path getBlockCommitStatusByHash          [lock-free precheck]
            (s.mu.Lock)                                          ← double-check afterwards
            |-> 1. core/minorblockchain.go:1048  InsertChainForDeposits (force=true)
            |         (m.chainmu.Lock)
            |         |-> WriteBlockWithState :914   (m.mu.Lock)
            |         |-> updateTip :158             (m.mu.Lock)
            |         (m.chainmu.Unlock)
            |         |-> post-insert reads confirmedHeaderTip (m.mu.Lock)
            |-> 2. conn.BroadcastXshardTxList                     [no lock]
            |         [error] s.setHead :601 -> MinorBlockChain.SetHead :318 (chainmu.Lock -> m.mu.Lock)
            |-> 3. GetShardStats :1214 -> getBlockCountByHeight (m.mu.RLock)
            |         [error] s.setHead -> SetHead :318 (chainmu.Lock -> m.mu.Lock)
            |-> 4. conn.SendMinorBlockHeaderToMaster              [no lock]
            |         [error] s.setHead -> SetHead :318 (chainmu.Lock -> m.mu.Lock)
            |-> 5. core/minorblockchain_addon.go:1931  CommitMinorBlockByHash
            |         (m.mu.Lock; checks HasBlock then writes marker; returns false→ErrBodyDeleted)
            |-> 6. broadcastNewTip :591
                      |-> GetRootTip :1434 (m.mu.RLock)
```

> `s.setHead` (shard layer, api_backend.go:601) calls `MinorBlockChain.SetHead` (chainmu→mu).
> On error branches setHead is invoked while holding **only s.mu**, not m.chainmu/m.mu, so SetHead re-acquiring both is not re-entrant.

---

## 5. AddBlockListForSync

```
cluster/slave/api_backend.go  SlaveBackend.AddBlockListForSync
|   (HasCommittedBlock filters already-committed blocks)
|-> cluster/shard/api_backend.go:465  ShardBackend.AddBlockListForSync
      (s.mu.Lock)
      |-> [per block] InsertChainForDeposits (force=true) :1048
      |         (m.chainmu.Lock)
      |         |-> WriteBlockWithState :914  (m.mu.Lock)
      |         |-> updateTip :158            (m.mu.Lock)
      |         (m.chainmu.Unlock)
      |         |-> post-insert reads confirmedHeaderTip (m.mu.Lock)
      |-> conn.BatchBroadcastXshardTxList                        [no lock]
      |-> conn.SendMinorBlockHeaderListToMaster                  [no lock]
      |-> [per header] CommitMinorBlockByHash :1931
                (m.mu.Lock; returns false→ErrBodyDeleted, aborts the whole batch)
```

---

## 6. CreateBlockToMine

```
cluster/shard/api_backend.go:640  ShardBackend.CreateBlockToMine
|-> core/minorblockchain_addon.go:898  MinorBlockChain.CreateBlockToMine
      |-> 1. m.mu.Lock :937  snapshot rootTip / currentBlock; m.mu.Unlock :940
      |-> 2. getEvmStateByBlock :1692
                (m.mu.Lock — only when block == CurrentBlock)
```

---

## 7. CheckMinorBlocksInRoot

```
cluster/slave/api_backend.go  SlaveBackend.CheckMinorBlocksInRoot
|-> cluster/shard/api_backend.go:668  ShardBackend.CheckMinorBlock
      |-> core/minorblockchain.go:1048  InsertChainForDeposits (force=true)
            (m.chainmu.Lock)
            |-> WriteBlockWithState :914  (m.mu.Lock)
            |-> updateTip :158            (m.mu.Lock)
            (m.chainmu.Unlock)
            |-> post-insert reads confirmedHeaderTip (m.mu.Lock)
```

---

## Lock Acquisition Order

All call paths acquire locks in the same order — no reverse acquisition, no AB-BA deadlock.

```
s.mu  (ShardBackend)
  └─ m.chainmu  (MinorBlockChain insertion lock)
       └─ m.mu  (MinorBlockChain state lock)
```

> **Note on AddRootBlock:** holds `m.chainmu` for the whole function (deferred at 1098). `m.mu` is explicitly
> unlocked at :1138, and only afterwards does it call `Reset` (which uses the unlocked `setHead`, taking neither
> chainmu nor mu) and `reWriteBlockIndexTo` (which takes only m.mu). Hence there is no re-entrant acquisition of
> chainmu/mu.
>
> **Error-branch setHead:** `ShardBackend.setHead` is only invoked while holding `s.mu`. Its inner
> `MinorBlockChain.SetHead` acquires chainmu→mu in the same order as the main path, so it is not re-entrant.

cluster/shard/api_backend.go
blockchaindevsh: PR changed `HasBlock` to body-only semantics, but did not update the existing `NewMinorBlock` callsite. This unchanged callsite now skips blocks that have a body but are not committed yet. Previously it'll retry.

core/minorblockchain_addon.go
blockchaindevsh: `SetHead` no longer serializes with `InsertChainForDeposits`. `insertChain` is documented to run with `chainmu` held, and the existing import path holds `chainmu` for the entire import to prevent canonical-head changes while historical state is being imported. Before this PR, external `SetHead` used the same mutex, so rewind/delete operations could not interleave with import. After switching `SetHead` to only `m.mu`, a rewind can delete block bodies/commit markers and move currentBlock while insertChain is still running. Please keep `SetHead` serialized with `chainmu` as before, and use `m.mu` only inside that critical section if needed for body/commit-marker consistency.

cluster/sync/minor_task.go
```diff
@@ -226,5 +230,9 @@ func (m *minorChainTask) findAncestor(bc blockchain) (*types.MinorBlockHeader, e
 			break
 		}
 	}
+
+	if !minorBlockCommitted(bc, bestAncestor.Hash()) {
```
blockchaindevsh: `bestAncestor` can still be `nil` if the remote minor chain has no committed local ancestor in `[start, end]`.

cluster/sync/minor_task.go
```diff
@@ -174,7 +165,20 @@ func (m *minorChainTask) downloadBlockHeaderListAndCheck(height, skip, limit uin
...
 
-func (m *minorChainTask) findAncestor(bc blockchain) (*types.MinorBlockHeader, error) {
+func minorBlockCommitted(bc *core.MinorBlockChain, h common.Hash) bool {
+	return bc.HasBlock(h) && bc.IsMinorBlockCommittedByHash(h)
+}
+
+func (m *minorChainTask) findAncestor(b blockchain) (*types.MinorBlockHeader, error) {
+	bc, ok := b.(*core.MinorBlockChain)
+	if !ok {
+		return nil, errors.New("Invalid blockchain type for minor chain task")
```
blockchaindevsh: This breaks the `blockchain` abstraction and existing tests/mocks. Existing cluster/sync/minor_task_test.go uses mockblockchain. With this change, those tests would return Invalid blockchain type for minor chain task once they get past needSkip. If commit status is needed here, prefer extending the interface or making the mock implement the needed behavior.

core/minorblockchain_addon.go
ping-ke: I'd like to clarify the locking design:

While `insertChain` runs under `chainmu`, the critical operations that modify canonical chain state (writing block bodies, updating `currentBlock`, writing commit markers) all happen inside `WriteBlockWithState`, which holds `m.mu` (line 896-897).

Similarly, `SetHead` also holds `m.mu` (line 312) when it deletes blocks and updates `currentBlock`. Therefore, these two critical sections are serialized via `m.mu`.

This design matches geth's approach:
- geth's `SetHead` uses `bc.mu`
- geth's `insertChain` uses `bc.chainmu`  
- geth's `writeBlockWithState` uses `bc.mu` for the actual writes

The role of `chainmu` is to serialize the overall import flow (prevent concurrent imports), while `m.mu` protects the canonical chain state consistency. The two locks work together to prevent the race you mentioned.

core/minorblockchain_addon.go
```diff
@@ -1099,40 +1104,51 @@ func (m *MinorBlockChain) AddRootBlock(rBlock *types.RootBlock) (bool, error) {
...
-			m.genesisBlock = newGenesis
-			log.Warn(m.logInfo+" ready to reset genesis", "number", m.genesisBlock.Number(), "hash", m.genesisBlock.Hash().String())
-			if err := m.Reset(); err != nil {
-				return false, err
-			}
-			break
-		}
+	// Rewind currentBlock until it is on the same root chain as rootTip.
+	// Stop at genesis (NumberU64==0); genesis reset is handled after Unlock
+	// because Reset() acquires chainmu and must not run under m.mu.
```
syntrust: Now Reset() → ResetWithGenesisBlock → SetHead(0) acquires m.mu, not chainmu.

cluster/shard/api_backend.go
```diff
@@ -296,12 +296,12 @@ func (s *ShardBackend) NewMinorBlock(peerId string, block *types.MinorBlock) (er
 	if s.mBPool.getBlockInPool(mHash) {
 		return
 	}
-	if s.MinorBlockChain.HasBlock(block.Hash()) {
+	if s.MinorBlockChain.HasCommittedBlock(block.Hash()) {
```
qzhodl, line 312: Switching this guard to HasCommittedBlock is not sufficient for body=yes, marker=no. The same body reaches ValidateBlock(block, false) at lines 328-330, which returns ErrKnownBlock via HasBlockAndState and exits before AddMinorBlock reaches its force=true recovery path. Please bypass this known-block pre-validation for an uncommitted body, or validate it with force=true, and add a NewMinorBlock retry test.

core/minorblockchain_addon.go
```diff
@@ -1099,40 +1104,51 @@ func (m *MinorBlockChain) AddRootBlock(rBlock *types.RootBlock) (bool, error) {
...
+	for m.CurrentBlock().NumberU64() != 0 &&
+		!m.isSameRootChain(m.rootTip, m.GetRootBlockByHash(m.CurrentBlock().PrevRootBlockHash())) {
 		preBlock := m.GetMinorBlock(m.CurrentBlock().ParentHash())
 		log.Warn(m.logInfo+" ready to set currentHeader", "height", preBlock.Number(), "hash", preBlock.Hash().TerminalString())
 		m.currentBlock.Store(preBlock)
 	}
+	needGenesisReset := m.CurrentBlock().NumberU64() == 0 &&
+		!m.isSameRootChain(m.rootTip, m.GetRootBlockByHash(m.CurrentBlock().PrevRootBlockHash()))
+	snapshotRootTip := m.rootTip
+	m.mu.Unlock()
```
qzhodl: This unlock publishes the new rootTip/currentBlock before reWriteBlockIndexTo updates the canonical/head/transaction indexes and before currentEvmState is updated at line 1157. A reader can acquire m.mu in between and observe currentBlock=newBlock while State or GetRootChainStakes still uses the old state; the later currentEvmState assignment is also outside m.mu. Please make the tip/index/state transition atomic instead of publishing a partially completed switch.

core/minorblockchain.go
```diff
@@ -309,8 +309,8 @@ func (m *MinorBlockChain) loadLastState() error {
 // nodes after a fast sync).
 // already have locked
 func (m *MinorBlockChain) SetHead(head uint64) error {
-	m.chainmu.Lock()
-	defer m.chainmu.Unlock()
+	m.mu.Lock()
```
qzhodl, line 321: Serializing only the final DB writes with m.mu does not protect the assumptions made earlier in insertChain. For example, C11 can be validated against parent P10 and updateTip can return true; SetHead(9) can then delete P10 before WriteBlockWithState reacquires m.mu. WriteBlockWithState writes the C11 body/state and reorg(A9, C11) then fails because P10 is gone, leaving a partial body. The pruned-sidechain path also writes through WriteBlockWithoutState without m.mu. Please keep SetHead excluded by chainmu for the full import, then acquire m.mu in a consistent order, or revalidate the parent/head after acquiring m.mu before writing.

cluster/shard/api_backend_test.go
```diff
@@ -0,0 +1,432 @@
...
+// The fix changes HasBlock to check the block body in the DB directly
+// (rawdb.ReadMinorBlock), independent of the commit marker.
+//
+// Note: the race that produces marker=present, body=absent is still possible
+// (CommitMinorBlockByHash is still called outside m.chainmu). This test does NOT
+// assert that the bad state never occurs; it asserts that HasBlock handles it
+// correctly when it does.
+//
+// Run with: go test -race ./cluster/shard/... -run TestAddBlockListForSync_MarkerBodyConsistencyUnderRace
+func TestAddBlockListForSync_MarkerBodyConsistencyUnderRace(t *testing.T) {
```
syntrust: The test cannot achieve its purpose: 
`SetHead` always delete both body and marker -> `!hasBody && hasMarker` never happens -> `if sb.MinorBlockChain.HasBlock(blockAFork.Hash()) {` never reachable.

cluster/shard/api_backend_test.go
```diff
@@ -0,0 +1,432 @@
...
+
+// TestAddMinorBlock_MarkerBodyConsistencyUnderRace verifies the Bug 1 fix
+// for AddMinorBlock: when a concurrent SetHead deletes the block body, HasBlock
+// must return false, not true.
+//
+// Same semantics as TestAddBlockListForSync_MarkerBodyConsistencyUnderRace:
+// we do not assert the bad state never occurs; we assert HasBlock handles it.
+//
+// Run with: go test -race ./cluster/shard/... -run TestAddMinorBlock_MarkerBodyConsistencyUnderRace
+func TestAddMinorBlock_MarkerBodyConsistencyUnderRace(t *testing.T) {
```
syntrust: The test cannot achieve its purpose: 
`SetHead` always delete both body and marker -> `!hasBody && hasMarker` never happens -> `if sb.MinorBlockChain.HasBlock(blockAFork.Hash()) {` never reachable.

core/minorblockchain.go
```diff
@@ -1158,9 +1184,12 @@ func (m *MinorBlockChain) insertChain(chain []types.IBlock, verifySeals bool, is
 			return it.index, events, coalescedLogs, xShardList, err
 		}
 
-		if isCheckDB {
+		if force && m.HasBlock(mBlock.Hash()) {
```
syntrust: The `force && HasBlock → continue` statement breaks the delayed state import mechanism of `insertSidechain`, preventing the re-org sidechain from ever completing.

syntrust (dismissed):

cluster/shard/api_backend.go
```diff
@@ -433,32 +469,53 @@ func (s *ShardBackend) AddBlockListForSync(blockLst []*types.MinorBlock) error {
...
+		seenInBatch[blockHash] = struct{}{}
+
+		// Validate and add the block with force=true. If the body already exists,
+		// the block is re-executed to recover the xshard list without writing state.
+		_, xshardLst, err := s.MinorBlockChain.InsertChainForDeposits([]types.IBlock{block}, true)
 		if err != nil {
 			log.Error(s.logInfo+" Failed to add minor block", "err", err)
 			return err
 		}
 		if len(xshardLst) != 1 {
```
qzhodl: [P1] Delayed sidechain import can still stall permanently here. For example, with local tip 200 and a sync batch [S101...S201] whose parent state was pruned, this loop calls InsertChainForDeposits with [S101] only. insertSidechain stores the S101 body and returns zero x-shard lists because 101 < 200, so this error aborts before S102 is processed. A retry starts from the same uncommitted S101 and fails identically, while NewMinorBlock also rejects S102 because HasCommittedBlock(S101) is false. Please process the contiguous segment together and associate every replayed x-shard list with its block, or otherwise allow body-only entries to advance until state reconstruction can run.

core/minorblockchain_addon.go
```diff
@@ -1099,47 +1120,75 @@ func (m *MinorBlockChain) AddRootBlock(rBlock *types.RootBlock) (bool, error) {
...
```
qzhodl: [P1] The latest change still publishes this transition in two phases. currentBlock may already have been changed at lines 1119/1133, but this unlock occurs before reWriteBlockIndexTo updates the canonical indexes and currentEvmState. For example, during a C10 -> C8 rewind, a reader in this window can observe CurrentBlock()==C8 while State() still returns C10 state and GetBlockByNumber still follows the old index. The genesis path has the same gap after Reset() unlocks and before currentEvmState is assigned. Please keep the target block local and publish rootTip, indexes, currentBlock, and currentEvmState atomically under one m.mu critical section.

core/minorblockchain.go
```diff
@@ -463,9 +472,15 @@ func (m *MinorBlockChain) Reset() error {
...
+//
+// Calls the unlocked setHead: every path that reaches here already holds the
+// relevant locks (SetHead holds chainmu+mu; AddRootBlock's genesis reset holds
+// chainmu; loadLastState runs during single-threaded init or from within
+// setHead which already holds them). Calling the exported SetHead here would
+// re-acquire chainmu/mu and self-deadlock.
 func (m *MinorBlockChain) ResetWithGenesisBlock(genesis *types.MinorBlock) error {
 	// Dump the entire block chain and purge the caches
-	if err := m.SetHead(0); err != nil {
+	if err := m.setHead(0); err != nil {
```
qzhodl: [P2] This makes the exported Reset/ResetWithGenesisBlock path unlocked. A direct Reset() call reaches this setHead(0) without taking chainmu or m.mu, so it can delete bodies/markers and move currentBlock concurrently with InsertChainForDeposits or CommitMinorBlockByHash. The comment assumption that every caller already holds the locks is not true for the public Reset wrapper. Please keep a locking public entry point and use a separate private resetWithGenesisBlockLocked helper for callers such as AddRootBlock that already hold the locks.

core/minorblockchain_addon.go
ping-ke: https://github.com/QuarkChain/goquarkchain/issues/693 create issue to check “reading  rootTip, indexes, currentBlock, and currentEvmState without lock” issue.

PR #694, opened 2026-07-10
fix (part 1): serialize minor head rewind and root reorg, and avoid conflicts with import / commit.

ping-ke:
This is the first PR split out from [QuarkChain/goquarkchain#689](https://github.com/QuarkChain/goquarkchain/pull/689).
This PR does not change commit marker semantics.

### Issue / impact:

- `SetHead` / root reorg can conflict with minor block import / commit, so a body may be deleted while still being committed or referenced.

### Root cause:

- rewind / root reorg / import were not fully serialized by the same set of locks.

### Fix:

- Use the lock order `s.mu -> chainmu -> mu`, so rewind / root reorg / import are mutually exclusive.
- `setHead` validates the target state first, then deletes bodies, then publishes head/state together.

### Review focus:

- Any reverse lock order or re-entrant deadlock?
- If rewind returns an error, can it return after deleting bodies?

### Tests

Covered by existing chain reorg/head tests and follow-up race coverage in later PRs.

### DIFF
--- a/cluster/shard/api_backend.go
+++ b/cluster/shard/api_backend.go
@@ -194,6 +194,13 @@ func (s *ShardBackend) InitFromRootBlock(rBlock *types.RootBlock) error {
 }
 
 func (s *ShardBackend) AddRootBlock(rBlock *types.RootBlock) (switched bool, err error) {
+	// Serialize root-block handling against AddMinorBlock / AddBlockListForSync on
+	// the same shard, so their broadcast + master-report side effects cannot
+	// interleave with root reorg rollback decisions. The core-layer race
+	// (currentBlock/canonical-index rewrite vs. insertChain) is handled inside
+	// MinorBlockChain.AddRootBlock via chainmu.
+	s.mu.Lock()
+	defer s.mu.Unlock()
 	s.wg.Add(1)
 	defer s.wg.Done()
 	switched = false
--- a/cluster/shard/api_backend.go
+++ b/cluster/shard/api_backend.go
@@ -182,6 +182,8 @@ func (s *ShardBackend) GetCode(recipient account.Recipient, height *uint64) ([]b
 // ######################## root block Methods #########################
 // Either recover state from local db or create genesis state based on config
 func (s *ShardBackend) InitFromRootBlock(rBlock *types.RootBlock) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
 	s.wg.Add(1)
 	defer s.wg.Done()
 	if rBlock.Number() > s.genesisRootHeight {
@@ -536,6 +538,10 @@ func (s *ShardBackend) broadcastNewTip() (err error) {
 }
 
 func (s *ShardBackend) setHead(head uint64) {
+	// TODO: This compensation rewind deletes block bodies but may leave stale
+	// transaction lookup entries. Clean up the rolled-back blocks here instead
+	// of adding unconditional index deletion to MinorBlockChain.SetHead, which is
+	// also used by reorg and recovery paths.
 	if err := s.MinorBlockChain.SetHead(head); err != nil {
 		panic(err)
 	}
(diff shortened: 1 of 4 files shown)


qzhodl:
I think this PR is mixing two separate problems.

The shard/core serialization with s.mu/chainmu addresses the writer-writer race between SetHead/root reorg and minor block import/commit. That is the critical correctness issue, and it is worth fixing here.

The *WithoutHeadPublish refactor seems to target reader consistency, but it is still only a partial solution. It avoids publishing currentBlock early, but those helpers still mutate canonical indexes / tx indexes / bodies before the final currentBlock/currentEvmState publish. Since readers such as CurrentBlock() and GetBlockByNumber() do not take the same read lock, they can still observe a mixed snapshot, just in the opposite direction.

So I suggest keeping this PR focused on the writer-writer serialization issue, and tracking the reader snapshot problem separately with a clear locking/snapshot policy for all relevant readers.

qzhodl (dismissed):

core/minorblockchain_addon.go
```diff
@@ -1069,9 +1071,19 @@ func (m *MinorBlockChain) AddRootBlock(rBlock *types.RootBlock) (bool, error) {
...
+	// Keep root-tip selection and canonical rewrites mutually exclusive with
+	// insertChain. Lock order is chainmu -> mu.
+	m.chainmu.Lock()
+	defer m.chainmu.Unlock()
+
+	// No change to root tip. rootTip is guarded by mu (written under mu here and
+	// read by CreateBlockToMine / InitGenesisState), so read it under mu rather
+	// than relying on chainmu alone.
+	m.mu.Lock()
+	curRootTip := m.rootTip
```
qzhodl: This only reads rootTip. Why do we need to put a lock around it?

cluster/shard/api_backend.go
```diff
@@ -194,6 +194,13 @@ func (s *ShardBackend) InitFromRootBlock(rBlock *types.RootBlock) error {
 }
 
 func (s *ShardBackend) AddRootBlock(rBlock *types.RootBlock) (switched bool, err error) {
+	// Serialize root-block handling against AddMinorBlock / AddBlockListForSync on
+	// the same shard, so their broadcast + master-report side effects cannot
+	// interleave with root reorg rollback decisions. The core-layer race
+	// (currentBlock/canonical-index rewrite vs. insertChain) is handled inside
+	// MinorBlockChain.AddRootBlock via chainmu.
+	s.mu.Lock()
```
qzhodl, line 204: AddRootBlock is now serialized with AddMinorBlock/AddBlockListForSync via s.mu, but InitFromRootBlock can still rewrite rootTip/currentBlock/currentEvmState/canonical indexes without taking s.mu. This path is reachable for existing shards through CreateShards(..., forceInit=true) from MasterInfo. If that can run after the shard is active, it has the same write-write race. Please either guard InitFromRootBlock with s.mu as well, or document/enforce that it is startup-only.

qzhodl (dismissed):

blockchaindevsh (dismissed):
I don't see a concrete deadlock path in the current PR. The shard-level paths use `s.mu -> m.chainmu -> m.mu`, and the core paths consistently use `m.chainmu -> m.mu`. I also don't see core calling back into `ShardBackend` while holding `m.chainmu`/`m.mu`.

The remaining caveat is that `s.mu` is held across synchronous conn/RPC calls. Today's master handlers don't synchronously call back into the same shard path, so this does not appear to deadlock now, but future changes should preserve that invariant or move those side effects outside `s.mu`.

qzhodl (approved):

core/minorblockchain_addon.go
syntrust, line 474: Should the write operation to m.rootTip be protected in L474 and L490?

ping-ke: `InitFromRootBlock` is a one-time operation; the likelihood of a conflict is very low, and there is no significant impact. To minimize the scope of changes, I haven't modified it for now but have deferred the update to a future issue.

syntrust (approved):

PR #695, opened 2026-07-10
fix: separate block body and commit marker status

ping-ke:
This is the second PR split out from [QuarkChain/goquarkchain#689](https://github.com/QuarkChain/goquarkchain/pull/689), based on `fix/minor-chain-head-locking`.

### Issue / impact:

- Body/state in the DB does not mean x-shard broadcast and master header report are done.
- The old code mixed body present with committed, so after crash / retry it could not tell which blocks still needed commit side effects.
- During concurrent rewind, a marker could be written for a block whose body was already deleted.

### Root cause:

- `WriteBlockWithState` wrote the commit marker while writing body/state, marking the block committed too early.
- `HasBlock` meant both "does the body exist" and "is commit done".
- Marker write was not an atomic check-and-write with body existence.

### Fix:

- `WriteBlockWithState` / `insertSidechain` no longer auto-write the commit marker. The marker is written only by the shard layer after x-shard broadcast + master header report succeed.
- `HasBlock` now only means "body present"; add `HasCommittedBlock` for body + marker.
- `CommitMinorBlockByHash` checks under `m.mu` that the body still exists before writing the marker. If rewind already deleted the body, it returns `false`.
- sync / slave use `HasCommittedBlock` when deciding whether a block is complete, so body-only blocks are not treated as committed.
- Blocks with body/state but missing marker can be re-executed to recompute the x-shard list, then broadcast / reported / marked.
- Remove the obsolete `BLOCK_COMMITTING` placeholder state.

### Review focus:

- Should each call site use `HasBlock` or `HasCommittedBlock`?
- Is the marker written only after broadcast / report succeeds?
- Does `marker => body` hold on both commit and delete paths?

### Tests

Added/updated coverage for:

- commit marker is not written when the block body is missing
- `InsertChain` writes block body/state without automatically writing the commit marker
- shard API backend writes commit markers after `AddMinorBlock` / `AddBlockListForSync` propagation succeeds
- shard API backend skips already committed blocks and commits uncommitted blocks during sync import
- race-tagged shard API backend coverage for commit marker/body consistency
- sync test mocks now model committed minor blocks explicitly

### DIFF
--- a/cluster/shard/api_backend.go
+++ b/cluster/shard/api_backend.go
@@ -24,6 +24,7 @@ import (
 var (
 	EmptyErrTemplate                 = "empty result when call %s, params: %v\n"
 	AllowedFutureBlocksTimeBroadcast = 15
+	ErrBodyDeleted                   = errors.New("block body deleted before commit")
 )
 
 // Wrapper over master connection, used by synchronizer.
@@ -305,14 +306,20 @@ func (s *ShardBackend) NewMinorBlock(peerId string, block *types.MinorBlock) (er
 	if s.mBPool.getBlockInPool(mHash) {
 		return
 	}
-	if s.MinorBlockChain.HasBlock(block.Hash()) {
+	if s.MinorBlockChain.HasCommittedBlock(block.Hash()) {
 		log.Debug("add minor block, Known minor block", "branch", block.Branch(), "height", block.Number())
 		return
 	}
 
-	if !s.MinorBlockChain.HasBlock(block.ParentHash()) {
-		log.Debug("prarent block hash not be included", "parent hash: ", block.ParentHash().Hex())
-		return
+	if !s.MinorBlockChain.HasCommittedBlock(block.ParentHash()) {
+		if err := s.commitUncommittedAncestorsIfPresent(block.ParentHash()); err != nil {
+			log.Warn(s.logInfo+" failed to commit uncommitted ancestors", "parent", block.ParentHash().Hex(), "err", err)
+			return nil
+		}
+		if !s.MinorBlockChain.HasCommittedBlock(block.ParentHash()) {
+			log.Debug("prarent block hash not be included", "parent hash: ", block.ParentHash().Hex())
+			return
+		}
 	}
 
 	//Sanity check on timestamp and block height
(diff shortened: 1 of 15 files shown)


core/minorblockchain.go
blockchaindevsh: procFutureBlocks still imports matured future blocks through m.InsertChain directly. Since this PR removed commit-marker writes from core insertion, this path can now write body/state and advance the head without running shard side effects or writing the commit marker.

cluster/shard/api_backend.go
```diff
@@ -451,8 +494,8 @@ func (s *ShardBackend) AddBlockListForSync(blockLst []*types.MinorBlock) error {
 		if s.getBlockCommitStatusByHash(blockHash) == BLOCK_COMMITTED {
 			continue
 		}
-		//TODO:support BLOCK_COMMITTING
-		_, xshardLst, err := s.MinorBlockChain.InsertChainForDeposits([]types.IBlock{block}, false)
+		forceReplay := s.MinorBlockChain.HasBlockAndState(blockHash)
```
blockchaindevsh, line 512: AddBlockListForSync can commit a child while its parent is only body/state-present but not committed. ValidateBlock only requires HasBlockAndState(parent), so if a previous crash left parent body/state without the marker, this batch path can send side effects and commit the child, leaving HasCommittedBlock(child)==true while HasCommittedBlock(parent)==false. The batch path needs the same ancestor-recovery/parent-commit invariant as NewMinorBlock, but implemented without recursively taking s.mu.

ping-ke: Add more block list check. and also check `if !s.MinorBlockChain.HasCommittedBlock(blockLst[0].ParentHash())` before insert block list.

cluster/shard/api_backend.go
```diff
@@ -481,16 +484,28 @@ func (s *ShardBackend) AddBlockListForSync(blockLst []*types.MinorBlock) error {
...
+		if i > 0 && (block.NumberU64() != blockLst[i-1].NumberU64()+1 || block.ParentHash() != blockLst[i-1].Hash()) {
+			return fmt.Errorf("sync block list is not parent-first contiguous at index %d", i)
+		}
+	}
+	// The slave removes committed blocks before calling this method. Since
+	// commits are parent-first, that can only remove a committed prefix. After
+	// verifying the remaining list is contiguous, its first parent therefore
+	// establishes the commit boundary for the entire batch.
+	if !s.MinorBlockChain.HasCommittedBlock(blockLst[0].ParentHash()) {
+		return fmt.Errorf("%w: %s", errParentBlockNotCommitted, blockLst[0].ParentHash().Hex())
```
blockchaindevsh, line 500: `NewMinorBlock` does ancestor recovery(the `commitUncommittedAncestorsIfPresent` call), but `AddBlockListForSync` doesn't, does the caller of `AddBlockListForSync` guarantee a committed boundary?

